PR #263, opened 2014-04-03
add test dependencies to package build dependencies

dirk-thomas:


k-okada:
I'm not sure if this is our special case, or general problem, but sometimes on x86 and always on arm, the buildfirm can not find package only included in `test_depends`
https://github.com/ros-infrastructure/ros_buildfarm/issues/539

PR #32, opened 2014-04-03
add support for building dbg debs separate from main packages

tfoote:
I worked on this inside catkin a while back.  The patch below is now way out of date, but should capture most of the issues.  

The old ticket was ros/catkin#20

A few notes from that ticket: 
semi official way to do it by debian is here: http://wiki.debian.org/DebugPackage of course it doesn't just work.

Ok, i've gotten it working in my fork. To do this correctly we should see if we can get cmake/catkin to produce the install files. http://manpages.ubuntu.com/manpages/natty/man1/dh_install.1.html  -- referring to the patch below. 

Old ticket in ros trac: https://code.ros.org/trac/ros/ticket/3365

```

From b6843f81421c0645a04c910e73ce79a4bff1b6b1 Mon Sep 17 00:00:00 2001
From: Tully Foote <tfoote@willowgarage.com>
Date: Sat, 11 Feb 2012 00:37:12 -0800
Subject: [PATCH] attempt to fix issue #20

---
 bin/catkin-generate-debian |    5 +++++
 bin/em/control.em          |    6 ++++++
 bin/em/package.install.em  |    1 +
 bin/em/rules.cmake.em      |    4 ++++
 4 files changed, 16 insertions(+)
 create mode 100644 bin/em/package.install.em

diff --git a/bin/catkin-generate-debian b/bin/catkin-generate-debian
index a1daf6d..753c6db 100755
--- a/bin/catkin-generate-debian
+++ b/bin/catkin-generate-debian
@@ -222,6 +222,11 @@ def generate_deb(stack_yaml, repo_path, stamp, debian_distro, rosdep_db):
     expand('changelog', stack_yaml, source_dir, dest_dir, filetype=stack_yaml['Catkin-ChangelogType'])
     expand('rules', stack_yaml, source_dir, dest_dir, filetype=stack_yaml['Catkin-DebRulesType'])
     expand('copyright', stack_yaml, source_dir, dest_dir, filetype=stack_yaml['Catkin-CopyrightType'])
+    try:
+        shutil.copyfile(os.path.join(source_dir, 'package.install.em'), os.path.join(dest_dir, '%s.install'%stack_yaml['Package']))
+    except Exception, ex:
+        print ("Failed to copy %s.rosintall into debian directory."%stack_yaml['Package'], ex)
+

     #compat to quiet warnings, 7 .. lucid
     ofilename = os.path.join(dest_dir, 'compat')
diff --git a/bin/em/control.em b/bin/em/control.em
index b95b13b..eed146d 100644
--- a/bin/em/control.em
+++ b/bin/em/control.em
@@ -10,3 +10,9 @@ Package: @(Package)
 Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends}, @(','.join(DebDepends))
 Description: @(Description)
+
+Package: @(Package)-dbg
+Section: debug
+Architecture: any
+Depends: @(Package) (= ${binary:Version}), ${misc:Depends}
+Description: @(Description)
diff --git a/bin/em/package.install.em b/bin/em/package.install.em
new file mode 100644
index 0000000..a16cadd
--- /dev/null
+++ b/bin/em/package.install.em
@@ -0,0 +1 @@
+/opt/ros/fuerte
diff --git a/bin/em/rules.cmake.em b/bin/em/rules.cmake.em
index ee50aa2..9a9c44d 100755
--- a/bin/em/rules.cmake.em
+++ b/bin/em/rules.cmake.em
@@ -26,3 +26,7 @@ override_dh_auto_configure:
 override_dh_auto_test:
    echo -- Running tests. Even if one of them fails the build is not canceled.
    dh_auto_test || true
+
+
+override_dh_strip:
+        dh_strip --dbg-package=@(Package)-dbg
-- 
1.7.10
```


vrabaud:
you still need to rebase to get https://github.com/ros-infrastructure/bloom/pull/371 in

tfoote:
This also needs a proper install file using \@InstallationPrefix is not clean and will miss cases.

nayomal:
Folks, is there more work to be done before this can be merged in? I'm guessing Tully's previous comment about needing install files was for supporting multiple debian packages (-dev, -doc etc) as builds that generate a single binary package typically don't need them.

tfoote:
This is building multiple packages you get foo and foo-dbg. My recollection from my testing was that the install file was required for the foo package to be properly built.

wjwwood:
@nayomal I think it still applies, but the biggest thing here is testing to make sure it doesn't break the farm. This change is particularly worrisome: 

https://github.com/ros-infrastructure/bloom/pull/32/files#diff-6986f9f31d980bb04856316ad94d707dR5

Changing the deb helper version has been tricky in the past.

nayomal:
Tully, yes you're right. That would make it a multi-package build. However we don't necessarily need *.install files if we just have a single package with all the content (ie: only a foo package without foo-dev, foo-doc etc) and just have an accompanying foo-dbg package. debhelper can simply look at the foo package and generate the foo-dbg package provided we have a package definition in the control file for foo-dbg and an "override_dh_strip" target in the rules file that defines how to generate the foo-dbg package. At least this has been my experience when adding debug packages to existing builds (including roscpp which we've been patching fairly frequently to generate debug symbols when developers ask for them). So I believe we can introduce debug packages without putting in further changes to make use of .install files.

William, it looks like things have improved considerably with respect to debhelper compatibility since this merge request was originally submitted. Nowadays the bloom debian generator sets the debhelper compatibility to 7 only if it's generating the build files for Ubuntu Oneiric. Otherwise it defaults to 9. Looking at the current Indigo packages, at the time of writing only 43 of them use 7, 1821 use 9. So most packages have already transitioned over to the latest stable compatibility level. Though I'm unsure how those 43 packages still on version 7 relate to such an old distribution as Oneiric as Indigo is officially supported only on Saucy and Trusty.

It would be nice if we could introduce debug packages in a more controlled manner (on a per ROS package basis) rather than enabling them for all ROS packages at once. This would help address your concern about enabling debug package builds for all of ROS negatively affecting the build farm by allowing us to enable debug symbols for one package at a time or for groups of packages at a time. It also seems wrong to provide an empty debug package if there are no debug symbols to offer. For example take a ROS package that provides a pure Python library. There are no native binaries in such a package and thus no debug symbols. Turning on debug symbols for those packages that don't have binary content would also use up a lot of resources on the build farm unnecessarily. When there are multiple ROS and Ubuntu distributions to contend with, a few seconds spent running strip and creating -dbg packages for each ROS package will most likely add up to hours of extra build time.

I think the above can be addressed if we add a flag to the package.xml file (something like a <debuginfo> tag under <export> with either a boolean attribute or boolean text value) that if present and set to true enables generation of a debug package. The templates for generating the debian build files could then insert the changes necessary to generate a debug package (this can certainly be done in templating modules like Jinja, bloom uses empy which I don't have experience with, but I would expect it to be doable assuming empy is reasonably feature rich). Is it feasible to modify the package.xml spec to introduce a new tag to accommodate this proposal?

Sorry for the long-winded reply. There was a fair bit to put into writing.

tfoote:
If you have a way to write the control file which does not require the .install file please update this PR to do so and we can then test it. 

I agree that building empty dbg packages is suboptimal, however I would not consider it a blocker. It is feasible to modify the package.xml spec, but it's a lot of work and something we try to keep very stable. A very tight corollary would be the `<architecture_independent/>` flag which is already in use and I think bloom could use for a conditional response for which template to use in this case.

With respect to the upgrade of the dh if we can test it thoroughly we can probably make the change. Can you share how you're querying which packages are built by which version?

nayomal:
Thanks for the feedback Tully! I'll submit the update to the PR as soon as I can (most likely tomorrow or the day after).

Good to hear that we can consider making a change to the package.xml spec at a later date to streamline the builds and get rid of false debug packages. And yes, something like `<architecture_independent/>` is exactly what I meant.

The approach I took to find the debhelper version used by each package was naive, but it's effective. I grabbed the source for all ROS Indigo packages and looked at the compat files. You can do this by following the instructions below:
 - Add the debian source repository for ROS to APT if you don't already have it:<br/>
   `sudo sh -c 'echo "deb-src http://packages.ros.org/ros/ubuntu trusty main" > /etc/apt/sources.list.d/ros-sources.list'`
 - Update APT:<br/>
   `sudo apt-get update`
 - Download all sources for ROS Indigo to a temporary directory (this step takes a while, around 10-15 minutes for me, but will vary depending on your geographic location):<br/>
   `mkdir -p /tmp/ros-sources`<br/>
   `cd /tmp/ros-sources`<br/>
   `for p in $(apt-cache search ros-indigo | cut -d " " -f 1); do apt-get source $p; done`
 - Query packages using compat level 7:<br/>
   `dh_compat=7`<br/>
  ` find /tmp/ros-sources -type f -a -name "compat" -exec grep -l ${dh_compat} {} \; | awk -F/ '{print $(NF-2)}'`
 - To query those using version 9 set `dh_compat` to 9 and repeat the above command. Optionally you can pipe the output through `wc -l` to get the package counts for each version.

nayomal:
Tully, the update is posted in PR #374. I had to post a new PR as it's off my own fork. Please let me know if you run into any issues with the submission.

NikolausDemmel:
Since it was mentioned, above, has there been any work towards splitting binary and development packages (maybe also documentation)? If I understood the discussion correctly, it was at least envisioned that with catkin/bloom this should be possible at some point, right? Is there any conceptual documentation or discussion on that topic, or does someone have an idea on what would be needed to get this working (on top of building debug packages as probposed in #374)?

nayomal:
Yes, you'll get Debian build files that'll generate the debug package if you re-bloom using a version of bloom with this set of patches. One way to test it locally is using virtualenv. Steps to do this are as follows:
 - Create virtualenv:<br/>
   `$ virtualenv <path to bloom test virtualenv>`
 - Activate virtualenv (all subsequent commands should be run from this shell that has the virtualenv activated):<br/>
  ` $ . <path to bloom test virtualenv>/bin/activate`
 - Clone the fork in this pull request and checkout the branch with the relevant commits or apply all patches in the pull request on top of the latest upstream clone of bloom.
 - Install bloom in the virtualenv:<br/>
  `$ cd <path to patched clone of bloom>`<br/>
  `$ python setup.py install`
 - Run bloom-generate on a clone of the release repo to create the Debian files and then build using your preferred debian package build tools

I'm not sure when this is expected to be pulled in to the mainline. Originally we were going to test the patches on the build farm and merge into master if things look good there. I think @wjwwood or @tfoote may know whether this is still the plan.

tfoote:
A few updates on this. Debian has rolled out dbgsym packages by default as of Stretch. 
https://lists.debian.org/debian-devel/2015/12/msg00262.html

That version is in Ubuntu as of xenial: https://packages.ubuntu.com/xenial/debhelper however Ubuntu has patched the default behavior not to build the dbgsym package: http://changelogs.ubuntu.com/changelogs/pool/main/d/debhelper/debhelper_9.20160115ubuntu3/changelog

```
  * dh-strip: Adjust reversion of commit f1a803456 to account for the recent
    ENABLE_DDEBS â†’ ENABLE_DBGSYM rename. This actually disables -dbgsym
    generation by dh_strip by default again. (LP: #1535949)
```
@mikaelarguedas and I noticed that the dbgsym packages are much larger. For example roscpp is ~500kB but the dbgsym package is almost 10MB. (See http://packages.ros.org/ros/ubuntu/pool/main/r/ros-lunar-roscpp/) 

image

And comparing to the older builds and on other platforms we're surprised that the regular library/package seems to be the same size. Where we expected the stretch version to be smaller with the separate debug symbols package. 

Related it appears that debian is hosting the dbgsym packages in a separate apt repository that needs to be explicitly turned on: https://michael-prokop.at/blog/2017/05/26/the-newinstretch-game-dbgsym-packages-in-debianstretch/ This seems understandable since the dbgsym packages appear to be so large. OpenCV3 is 13MB for the deb and 67 for the dbgsym. http://packages.ros.org/ros/ubuntu/pool/main/r/ros-lunar-opencv3/

From testing we've confirmed that the debug symbols appear to be stripped from the primary deb on stretch. And gdb can load the debug symbols installed `Reading symbols from ./talker...Reading symbols from /usr/lib/debug/.build-id/17/07bba45a800181e644ac26d7d763db0654698b.debug...done.`

It also appears that the debug symbols are not present in the rest of the debian packages.

Looking at the debian hosting it makes sense to segregate the debug symbols from the default installation. Instead of pushing to a completely separate url we think it would make sense to add a new `debug` component to the debian repos. And when importing we should push the regular packages to `main` and the dbgsym packages to `debug`. They're already labeled by `Section` 'debug' and 'msic'

They also have a control element `Auto-Built-Package: debug-symbols` that might be usable to for filtering. And then we need to add a small section to the installation/debuggin instructions to add the `debug` component to the apt sources to get the debug symbols. This will allow us to use the same hosting but not have users accidentally downloading all the large dbgsym packages by default when they use a ros-lunar-something-* regex to install naively. Doing this will also require updating the import rules to do multiple components in parallel.

@j-rivero @nuclearsandwich FYI

mikepurvis:
This is consistent with our experience— the stripped symbols are 1-2 orders of magnitude larger than the binaries from which they come. Note one thing which we hit shipping symbols from our bundle builds is that the buildids are one global namespace, so if you release the same version of roscpp as `ros-kinetic-roscpp` and `ros-lunar-roscpp`, you may find that because of reproducible builds, you get identical buildids for some of the underlying binaries, especially ones that are trivial, and that may mean that the two dbgsym packages will conflict.

It's a bit of a stretch that you'll run into this, but it's something that's caused us a fair bit of a headache (we resolved it by archiving each individual stripped symbol file in its own deb, rather than a big blob of them together).

j-rivero:
Your proposal makes sense to me. I did not notice that Ubuntu has disabled the `-dbgsym` generation which is not a trivial decision since most of us maintaining software in Debian have migrated our packages to it and Ubuntu is getting the packages from Debian.

If I'm not wrong, Xenial and [Zesty](http://changelogs.ubuntu.com/changelogs/pool/main/d/debhelper/debhelper_9.20160115ubuntu3/changelog) has the patch to avoid `-dbgsym` creation (probably to gain some time for patching launchpad buildfarm). The support seems to be back in [Artful](http://changelogs.ubuntu.com/changelogs/pool/main/d/debhelper/debhelper_10.7.2ubuntu2/changelog).

tfoote:
Well maybe we should just be patient and focus on making sure we segregate the repos cleanly and we'll have the dbgsym packages once they're turned on by Ubuntu. I like doing less customization and sticking with the platform defaults.

From that changelog file I also see a comment about changing the default compression level that might help the size in the future.

j-rivero:
> Looking at the debian hosting it makes sense to segregate the debug symbols from the default installation. Instead of pushing to a completely separate url we think it would make sense to add a new debug component to the debian repos. And when importing we should push the regular packages to main and the dbgsym packages to debug. They're already labeled by Section 'debug' and 'msic'

It could be that Debian is using a separate repository for debug packages in order to avoid the mirrors to sync them and require larges amount of disk space. I'm not sure if the existing tools let us to sync with exclusions on Sections.

nuclearsandwich:
@tfoote may I rebase this against the reorganized files in 0.6.0 so we can start prepping this for Melodic?

tfoote:
I think letting the default behavior take over adding this specific logic makes sense. In particular since the curent state is incomplete in that we'd need to generate the contents of `@(Package).install` more correctly before we went forward with this PR. And then when it became default we'd likely want to revert this.

As for the hosting of the repos, I'd suggest that we not try to do too many fancy redirects and just host the debug files in our main repo. Looking at our current repository the low hanging fruit for reducing size is often removing resources (meshes/images/logs usually) rather than code which will be unaffected by the debug symbols. And disk space keeps getting cheaper so unless until we see a major issue we can just keep everything together. And this will keep things simpler for users who want to use the debug symbols.

So long story short. I suggest close this, let the default debug packages take over as of Artful, and don't try to segregate/redirect the debug symbol packages.

PR #358, opened 2015-02-03
assert that all dependencies are resolved to the default installer

wjwwood:
This pull request adds a check to `bloom` which asserts that all key's resolved installer matches the OS's default installer. This check is done for Debian/Ubuntu and RPM generators.

The error looks like this:

```
Pre-verifying Debian dependency keys...
Running 'rosdep update'...
Key 'python-percol' resolved to '['percol']' with installer 'pip', which does not match the default installer 'apt'.
Failed to resolve python-percol on ubuntu:saucy with: Error running generator: The Debian generator does not support dependencies which are installed with the 'pip' installer.
python-percol is depended on by these packages: ['jsk_tools']
<== Failed
Key 'python-percol' resolved to '['percol']' with installer 'pip', which does not match the default installer 'apt'.
Failed to resolve python-percol on ubuntu:trusty with: Error running generator: The Debian generator does not support dependencies which are installed with the 'pip' installer.
python-percol is depended on by these packages: ['jsk_tools']
<== Failed
Some of the dependencies for packages in this repository could not be resolved by rosdep.
You can try to address the issues which appear above and try again if you wish.
Would you like to try again? [Y/n]?
```

@cottsay FYI, this might not be desirable for Fedora, but I guessed that it was and included it anyways.


wjwwood:
@dirk-thomas

bloom/generators/common.py
```diff
@@ -116,7 +136,7 @@ def resolve_rosdep_key(
     ros_distro = ros_distro or DEFAULT_ROS_DISTRO
     view = get_view(os_name, os_version, ros_distro)
     try:
-        return resolve_for_os(key, view, installer, os_name, os_version)
+        return resolve_more_for_os(key, view, installer, os_name, os_version)
```
dirk-thomas: Should the other returns (e.g. line 143, 157) return the same number of values?

wjwwood: Yeah, good catch, updated in d4f21c9.

dirk-thomas:
LGTM

PR #363, opened 2015-03-13
update help prompts for newer distro names

codebot:


wjwwood:
Technically Hydro is still valid. Can you include hydro in those, or consolidate all of those to pull from one place and update that?

wjwwood:
You would just define it in the root of a module and everywhere you need it you would import that module and use the symbol from the module.

The question is what module, but I think there are some candidates like the `bloom.util` module:

https://github.com/ros-infrastructure/bloom/blob/master/bloom/util.py

wjwwood:
At best you could make the storage private and provide a getter function, like this:

https://github.com/ros-infrastructure/bloom/blob/master/bloom/util.py#L285-L287

With the "private" storage here:

https://github.com/ros-infrastructure/bloom/blob/master/bloom/util.py#L263

PR #366, opened 2015-05-08
Extend debian generator to split pkg description into <synopsis, long description>

gavanderhoorn:
Small change that formats the text in the `Description` field of the `control` file such that the first sentence will be treated as the synopsis (or short description) by `apt` and related tools. The rest of the package description is added as a single, large paragraph (as it is now).

All nose tests still pass, and I've added some new ones.

A more fancy version of `format_description()` could perform a proper conversion of embedded newlines or `<p>` tags in package descriptions, but I thought to start with this first.

Context: [Support for 'short description' in package.xml / Bloom?](http://answers.ros.org/question/112886/support-for-short-description-in-packagexml-bloom/).

### DIFF
--- a/bloom/generators/debian/generator.py
+++ b/bloom/generators/debian/generator.py
@@ -207,6 +207,29 @@ def format_depends(depends, resolved_deps):
     return formatted
 
 
+def format_description(value):
+    """
+    Format proper <synopsis, long desc> string following Debian control file
+    formatting rules. Treat first line in given string as synopsis, everything
+    else as a single, large paragraph.
+
+    Future extensions of this function could convert embedded newlines and / or
+    html into paragraphs in the Description field.
+
+    https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Description
+    """
+    value = debianize_string(value)
+    # NOTE: bit naive, only works for 'properly formatted' pkg descriptions (ie:
+    #       'Text. Text'). Extra space to avoid splitting on arbitrary sequences
+    #       of characters broken up by dots (version nrs fi).
+    parts = value.split('. ', 1)
+    if len(parts) == 1 or len(parts[1]) == 0:
+        # most likely single line description
+        return value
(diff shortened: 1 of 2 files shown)


bloom/generators/debian/generator.py
```diff
@@ -207,6 +207,26 @@ def format_depends(depends, resolved_deps):
...
+    formatting rules. Treat first line in given string as synopsis, everything
+    else as a single, large paragraph.
+
+    Future extensions of this function could convert embedded newlines and / or
+    html into paragraphs in the Description field.
+
+    https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Description
+    """
+    value = debianize_string(value)
+    parts = value.split('.', 1)
```
wjwwood: This should probably split on `'. '` to avoid splitting when something like a version number is used in the description.

wjwwood:
Other than the comment looks good to me.

bloom/generators/debian/generator.py
gavanderhoorn: Yes, true.

Splitting text correctly on sentence boundaries is hard to get robust though, so I opted for this naive way of doing it with just `split(..)`. It will only work correctly for properly formatted package descriptions then though.

gavanderhoorn: I'll see if I can get this a bit more robust. Notoriously hard though, so I don't have much hope (without relying on natural language parsers).

gavanderhoorn:
@wjwwood: I've updated the pattern used with `split()`. Turns out splitting a text into sentences is rather hard to make robust (abbreviations, punctuation marks, different languages, etc) without adding some form of natural language processing. I guess this is partly why the 'old' pkg manifests had an explicit *short description* field?

If the `description` element is not formatted as expected, the proposed change should result in the old behaviour (ie: all text in the synopsis).

wjwwood:
Yeah, I think this is a good enough. Thanks for looking into it for me!

PR #368, opened 2015-06-17
Add linebreaks to the generated README markdown

mikepurvis:
There's some disagreement about list formatting, as [discussed here](http://spec.commonmark.org/0.20/#lists). In a nutshell, Bitbucket follows the Markdown.pl style, whereas Github follows CommonMark.

This change makes the output render more nicely on both platforms.


wjwwood:
Sure, np.

PR #369, opened 2015-06-17
Fix typo in function name.

mikepurvis:


wjwwood:
Ah, thanks. With my editor, when I misspell something, it just auto completes to the misspelling every where else.

Issue #327, opened 2014-11-22
Default release build flags for packages


wjwwood:
This currently applies to Ubuntu/Debian but could also affect others by way of common policy. Currently we do not explicitly set any release related compiler or CMake flags like `-DNDEBUG`, `-O3`, or `-DCMAKE_BUILD_TYPE=Release`, the question is should we and if so which ones?

tl;dr:

As I see it we can choose one of these options:
- Add `-DCMAKE_BUILD_TYPE=Release` (effectively `-O3 -DNDEBUG`) to the bloom template
  - This will (through attrition) make it so all packages have these flags
  - They must explicitly opt-out to return to the current settings
- Same as above but only `-NDEBUG`
- Again the same but only `-O3`
- Or do nothing in bloom, but write up how to explicitly opt-into these flags
  - This allows developers who care to opt-in to potentially breaking compiler flags

I am leaning towards the third option, but I would like some feedback @dirk-thomas @tfoote @mikeferguson @isucan @davetcoleman @mintar @tlind @j-rivero

---

Related tickets:
- https://github.com/ros-visualization/rviz/issues/775
- https://github.com/ros-planning/moveit_core/pull/204
- https://github.com/ros-infrastructure/bloom/pull/278

---

Currently we pass some non-release related flags through `deb_helper`:

https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/rules.em#L27-30

The `CATKIN_BUILD_BINARY_PACKAGE` CMake argument is how we disable installing of `setup.bash`.

My understanding (given the research I have done into things) is that `deb_helper` inserts default values into the build, for both CMake and the compiler and some of those defaults come from `dpkg-buildflags`. This happens automatically as of debhelper version 9 (which bloom is already using, see: https://github.com/ros-infrastructure/bloom/pull/278), see: https://wiki.debian.org/HardeningWalkthrough#Enabling_dpkg-buildflags_in_your_debian.2Frules_files

In accordance with this, `deb_helper` prepends our CMake variables with `-DCMAKE_BUILD_TYPE=None`. This (I think) prevents CMake from overriding by default the flags `deb_helper` passes to the compiler. We could override this by adding `-DCMAKE_BUILD_TYPE=Release` in the bloom rules template linked above, however, I think this potentially dangerous.

[...]

PR #374, opened 2015-09-24
Generate debian debug packages

nayomal:
This changeset makes bloom generate build files that create a debug symbols
debian package for all ROS packages. As discussed in pull request #32 a future
to-do is to enable building of debug symbols only for ROS packages that have
binary content.

### DIFF
--- a/bloom/generators/debian/templates/control.em
+++ b/bloom/generators/debian/templates/control.em
@@ -17,5 +17,5 @@ Package: @(Package)-dbg
 Architecture: any
 Section: debug
 Priority: extra
-Depends: @(Package) (= ${binary:Version}), ${shlibs:Depends}, ${misc:Depends}
+Depends: @(Package) (= ${binary:Version}), ${misc:Depends}
 Description: Debug symbols for @(Package)
--- a/bloom/generators/debian/templates/control.em
+++ b/bloom/generators/debian/templates/control.em
@@ -12,3 +12,10 @@ Depends: ${shlibs:Depends}, ${misc:Depends}, @(', '.join(Depends))
 @[if Conflicts]Conflicts: @(', '.join(Conflicts))@\n@[end if]@
 @[if Replaces]Replaces: @(', '.join(Replaces))@\n@[end if]@
 Description: @(Description)
+
+Package: @(Package)-dbg
+Architecture: any
+Section: debug
+Priority: extra
+Depends: @(Package) (= ${binary:Version}), ${shlibs:Depends}, ${misc:Depends}
+Description: Debug symbols for @(Package)
--- a/bloom/generators/debian/templates/rules.em
+++ b/bloom/generators/debian/templates/rules.em
@@ -30,8 +30,7 @@ override_dh_auto_configure:
 	dh_auto_configure -- \
 		-DCATKIN_BUILD_BINARY_PACKAGE="1" \
 		-DCMAKE_INSTALL_PREFIX="@(InstallationPrefix)" \
-		-DCMAKE_PREFIX_PATH="@(InstallationPrefix)" \
-		-DCMAKE_BUILD_TYPE=RelWithDebInfo
+		-DCMAKE_PREFIX_PATH="@(InstallationPrefix)"
 
 override_dh_auto_build:
 	# In case we're installing to a non-standard location, look for a setup.sh
--- a/bloom/generators/debian/templates/rules.em
+++ b/bloom/generators/debian/templates/rules.em
@@ -30,7 +30,8 @@ override_dh_auto_configure:
 	dh_auto_configure -- \
 		-DCATKIN_BUILD_BINARY_PACKAGE="1" \
 		-DCMAKE_INSTALL_PREFIX="@(InstallationPrefix)" \
-		-DCMAKE_PREFIX_PATH="@(InstallationPrefix)"
+		-DCMAKE_PREFIX_PATH="@(InstallationPrefix)" \
+		-DCMAKE_BUILD_TYPE=RelWithDebInfo
 
 override_dh_auto_build:
 	# In case we're installing to a non-standard location, look for a setup.sh
@@ -59,4 +60,7 @@ override_dh_auto_install:
 	# in the install tree that was dropped by catkin, and source it.  It will
 	# set things like CMAKE_PREFIX_PATH, PKG_CONFIG_PATH, and PYTHONPATH.
 	if [ -f "@(InstallationPrefix)/setup.sh" ]; then . "@(InstallationPrefix)/setup.sh"; fi && \
-	dh_auto_install
+	dh_auto_install --destdir=$(CURDIR)/debian/@(Package)
+
+override_dh_strip:
+	dh_strip --dbg-package=@(Package)-dbg


bloom/generators/debian/templates/rules.em
```diff
@@ -30,7 +30,8 @@ override_dh_auto_configure:
 	dh_auto_configure -- \
 		-DCATKIN_BUILD_BINARY_PACKAGE="1" \
 		-DCMAKE_INSTALL_PREFIX="@(InstallationPrefix)" \
-		-DCMAKE_PREFIX_PATH="@(InstallationPrefix)"
+		-DCMAKE_PREFIX_PATH="@(InstallationPrefix)" \
+		-DCMAKE_BUILD_TYPE=RelWithDebInfo
```
wjwwood: This is likely to have an impact on our packages. Currently we don't set `Release`, so `RelWithDebInfo` may do more than just add debug info. See: https://github.com/ros-infrastructure/bloom/issues/327

bloom/generators/debian/templates/rules.em
```diff
@@ -59,4 +60,7 @@ override_dh_auto_install:
 	# in the install tree that was dropped by catkin, and source it.  It will
 	# set things like CMAKE_PREFIX_PATH, PKG_CONFIG_PATH, and PYTHONPATH.
 	if [ -f "@(InstallationPrefix)/setup.sh" ]; then . "@(InstallationPrefix)/setup.sh"; fi && \
-	dh_auto_install
+	dh_auto_install --destdir=$(CURDIR)/debian/@(Package)
```
wjwwood: Why is this necessary?

tfoote:
Do we need to use the nostrip option as defined here: https://www.debian.org/doc/debian-policy/ch-source.html#s-debianrules-options ?

bloom/generators/debian/templates/rules.em
nayomal: Thanks for the reference! I have removed this and will push the update shortly. I did a quick test and the default build type still passes -g to the compiler so the debug symbols are generated.

bloom/generators/debian/templates/rules.em
nayomal: This is needed because dh_auto_install installs into debian/tmp instead of debian/@(Package) when we introduce a 2nd debian package (the -dbg package) to the control file. Asking dh_auto_install to install the content explicitly into debian/@(Package) using --destdir is often used in such build configurations as an alternative to using .install files.

nayomal:
I haven't used nostrip before, but I don't think we need it as it looks like it's provided for folks who want to quickly rebuild a package with unstripped binaries without needing to add an override target for dh_strip in the rules file. An example of it being used this way can be seen here:
https://wiki.debian.org/BuildingTutorial#Building_the_modified_package

I also looked at the dh_strip code and it simply exits before performing the strip operations if nostrip is defined as a debian build option.

bloom/generators/debian/templates/rules.em
wjwwood: Ok, I'm most uncertain/concerned about this part of the change because I'm unfamiliar with the underlying mechanism, but we can possibly test it out to make sure it doesn't break anything.

bloom/generators/debian/templates/rules.em
wjwwood: I think I remember that now. I was going to suggest just adding the `-g` option instead of setting the `CMAKE_BUILD_TYPE`, but it seems that this is not necessary.

bloom/generators/debian/templates/rules.em
nayomal: Thanks William, yes it would be great if we could test it out on the farm to see how well it fares. I came across this Debian wiki that partially describes this approach:
https://wiki.debian.org/DebugPackage

Please see the section titled "Debhelper >= 7"

wjwwood:
@tfoote what's the best way to test this? Can we do it on the test farm or should we wait until you get the new build farm deployed?

tfoote:
We can start by running it locally building and replacing debs on our development machines. Once we've verified that, we can make a new release to a gbp with it. RIght now our test farm and main farm use the same rosdistro so it would be most convenient to try out a leaf package. Maybe something like the serial lib would be good?  If we release it we can make sure to not let it out of shadow-fixed/testing.

bloom/generators/debian/templates/control.em
```diff
@@ -12,3 +12,10 @@ Depends: ${shlibs:Depends}, ${misc:Depends}, @(', '.join(Depends))
 @[if Conflicts]Conflicts: @(', '.join(Conflicts))@\n@[end if]@
 @[if Replaces]Replaces: @(', '.join(Replaces))@\n@[end if]@
 Description: @(Description)
+
+Package: @(Package)-dbg
+Architecture: any
+Section: debug
+Priority: extra
+Depends: @(Package) (= ${binary:Version}), ${shlibs:Depends}, ${misc:Depends}
```
j-rivero: Any reason to include shlibs in Depends? [Debian documentation](https://wiki.debian.org/DebugPackage) does not include it, probably because they will be pulled form the `$(Package)`.

bloom/generators/debian/templates/rules.em
j-rivero: From my experience as debian maintainer, the destdir is necessary when building -dbg packages, probably because we don't have a single package in the control file any more.

In debian I use `--destdir` with a relative path `debian/$(Package)`, which is the form that appears in the [documentation](https://wiki.debian.org/DebugPackage). @nayomal was the CURDIR strictly necessary for the integration with bloom?

j-rivero:
A part from my comments, the changes looks good to me from the debian packaging perspective. I can not think on an use case which could be broken by this PR, if we get one package correctly, the rest have minimal chances of fail AFAIK.

Large testing in the ROS buildfarm would be awesome, yes.

bloom/generators/debian/templates/control.em
nayomal: @j-rivero, thanks for the feedback! You're absolutely right, shlibs isn't needed for the dbg package. It must have crept in out of habit. I'll push an update shortly to remove it.

bloom/generators/debian/templates/rules.em
nayomal: No, CURDIR isn't strictly necessary. The dbg package will happily build even if a relative path is used. I used CURDIR for consistency with the rest of the rules (ie: the override_dh_shlibdeps target uses `$(CURDIR)/debian`). We can switch to a relative path if that's what is recommended. I've seen CURDIR used fairly often in rules files though.

j-rivero: If we are currently using it, I agree with you that makes sense to keep consistency in all the rules. No problem, keep it. Thanks!

NikolausDemmel:
What is the state of this? The need for debug packages just came up for us.

How can I test this locally? I guess I can just re-bloom one of my packages with a patched version of bloom including this change and then build the resulting debians for this package locally? Is there something I need to do to catkin / my package configuration to make sure the debug symbols are being generated, or are they already with the default build options?

tfoote:
If you could do some more testing building your package locally that would be great. We haven't had much time for testing this. We want to be careful with this since if we screw up it will significantly increase the burden on our maintainers. @NikolausDemmel if you could test it locally first and then also consider making a test release of a leaf package from this branch we could see how it works on the farm. 

I suspect that there will be a few places that we will need to expand package matching rules in the repository maintenance scripts that will be very hard to find w/o trying it out. 

You should not override the build flags. They should be set by the debbuilding process.

nuclearsandwich:
Since debug symbols are stripped and placed into -dbgsym packages by default on the current Debian and Ubuntu toolchains can this pull request be closed?

nayomal:
Closing this pull request as per suggestion from @nuclearsandwich.

nuclearsandwich:
@nayomal thanks for the contribution. Even though this PR didn't make merge the conversations and review it received improved our understanding and handling of dbgsym packages when they arrived via upstream.

nayomal:
My pleasure @nuclearsandwich :-)

PR #376, opened 2015-11-25
Fixing detection of rosdistro user fork when trying to create a pull request

asmodehn:
Also fixed when creating a rosdistro fork with a different name.

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -791,7 +791,8 @@ def open_pull_request(track, repository, distro, interactive, override_release_r
         try:
             repo_forks = gh.list_forks(base_org, base_repo)
             user_forks = [r for r in repo_forks if r.get('owner', {}).get('login', '') == gh.username]
-            head_repo = user_forks[0]  # github allows only 1 fork per org as far as I know. We just take the first one.
+            # github allows only 1 fork per org as far as I know. We just take the first one.
+            head_repo = user_forks[0] if user_forks else None
 
         except GithubException as exc:
             debug("Received GithubException while checking for fork: {exc}".format(**locals()))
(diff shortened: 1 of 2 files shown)


wjwwood:
I already have a fork of `ros/rosdistro`, but when I run this pr I get:

```
==> Checking on GitHub for a fork to make the pull request from...
Could not find a fork of ros/rosdistro on the wjwwood GitHub account.
Would you like to create one now?
Continue [Y/n]? y
==> Using this fork to make a pull request from: wjwwood/rosdistro
==> Cloning wjwwood/rosdistro...
==> mkdir -p rosdistro
==> git init
```

Anyways. It does work if I say yes, but it asks me every time. Even if I rerun with `-p` after answering `y` to the request to make a fork. So this doesn't look like the check is working correctly to me.

asmodehn:
This doesn't happen for me : 
```
==> Checking on GitHub for a fork to make the pull request from...
==> Using this fork to make a pull request from: asmodehn/rosdistro
```

I m thinking you might be hitting that exception handler:
```python
        except GithubException as exc:
            debug("Received GithubException while checking for fork: {exc}".format(**locals()))
            pass  # 404 or unauthorized, but unauthorized should have been caught above
```
and the debug log doesn't display anything ? 
It would be good to confirm this, if you can step through it, and find out the reason... Maybe we should display more information in that case ?

and then later on, this 
```python
                head_repo = gh.create_fork(base_org, base_repo)  # Will raise if not successful
```
selects automatically your already existing fork ? 
That would be an interesting behavior, although undocumented in the API docs, so we probably shouldn't rely on it.

asmodehn:
I did the previous test trying to release on yujinrobot/rosdistro from my fork asmodehn/rosdistro.
Maybe something slightly different happens when releasing on ros/rosdistro ? I didn't test releasing there yet.

asmodehn:
I believe I found the reason for that problem : the list_forks interface doesnt support paging, and there are a lot of forks of rosdistro. I ll fix that soon...

asmodehn:
Everything should work fine now. Let me know if there is anything else.

wjwwood:
I'll have a look at it asap.

wjwwood:
Finally got around to testing this. lgtm, thanks!

PR #380, opened 2016-03-05
add building of debian platforms to the default actions

tfoote:
They will fail gracefully for the older rosdistros which do not have debian platform targets.

As we discussed with @wjwwood [from notes](https://github.com/ros/rosdistro/pull/10621#issuecomment-192541582)

I'd like to parameterize the default actions template on the rosdistro, but that is a much more effort. It will work for now but we will reencounter this any time we add or remove a supported platform. And this will cause indigo and jade instances of bloom to recommend updating the tracks even though they don't need it. 

I've tested this with a Kinetic gbp and it generates the debian targets. And on an indigo track it updates the track and skips the debian targets.

### DIFF
--- a/bloom/config.py
+++ b/bloom/config.py
@@ -200,7 +200,9 @@ DEFAULT_TEMPLATE = {
         'git-bloom-generate -y rosrelease :{ros_distro}'
         ' --source upstream -i :{release_inc}',
         'git-bloom-generate -y rosdebian --prefix release/:{ros_distro}'
-        ' :{ros_distro} -i :{release_inc}',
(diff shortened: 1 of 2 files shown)


wjwwood:
lgtm

tfoote:
A test run generated an odd line here: https://github.com/ros-gbp/catkin-release/commit/3009cf1a0496253c1ddab92dd855b1667952fbec#diff-b9cba95a33d64aec1406327c4d43fa8dR1

wjwwood:
@tfoote thanks for pointing it out: https://github.com/ros-infrastructure/bloom/pull/381

PR #418, opened 2017-02-23
add a prompt in bloom to ask if the maintainer would like to enable pull request testing

tfoote:
This will hopefully help increase usage of PR testing.

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -527,6 +527,14 @@ def generate_ros_distro_diff(track, repository, distro, override_release_reposit
             info("Please enter information which points to the active development branch for this repository.")
             info("This information is used to run continuous integration jobs and for developers to checkout from.")
             source = get_repository_info_from_user('source', defaults)
+
+            if validate_github_url(source['url'], 'source'):
+                info("Since you are on github we can add a job to run your tests on each pull request."
+                     "If you would like to turn this on please see "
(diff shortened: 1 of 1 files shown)


bloom/commands/release.py
```diff
@@ -527,6 +527,14 @@ def get_repository_info_from_user(url_type, defaults=None):
             info("Please enter information which points to the active development branch for this repository.")
             info("This information is used to run continuous integration jobs and for developers to checkout from.")
             source = get_repository_info_from_user('source', defaults)
+
+            if validate_github_url(source['url'], 'source'):
+                info("Since you are on github we can add a job to run your tests on each pull request."
+                     "If you would like to turn this on please see "
+                     "http://wiki.ros.org/buildfarm/Pull%20request%20testing for more information. "
+                     "There is more setup required to setup the hooks correctly. ")
+                if maybe_continue(msg='Would you like to turn on pull request testing?', default='no'):
```
dirk-thomas: The current default value doesn't seem to work. When just hitting enter it repeats the question after printing the error:

> Reponse `no` was not recognized, please use one of y, Y, n, N.

wjwwood: https://github.com/ros-infrastructure/bloom/pull/420

Please open pr's for trivial things like this in the future.

wjwwood: Should be fixed in `0.5.26` which was just released.

PR #427, opened 2017-05-19
add missing 'unmaintaned' status

k-okada:
rep 141 defines developed, maintaned, unmaintaned and end-of-life

http://www.ros.org/reps/rep-0141.html#distribution-file


k-okada:
sorry for late, fixed to pass pep8 test

wjwwood:
Thanks!

PR #428, opened 2017-06-01
Add --debian-inc option to bloom-generate

jspricke:


nuclearsandwich:
@jspricke this pull request has been open for a while and doesn't appear to have rotted at all. Are you still interested in seeing it merged?

Nothing stands out as a possible issue to me.

The only potential reservations I have are:

* If we'd favor implementing a package revision option that is not distribution specific. But as each distribution has variation in the concept perhaps distro specific is best...

* Do we need to put any protections in to prevent maintainers from re-blooming an earlier debinc/pkgrel? While we can't stop folks from doing silly things on their own projects it would drive the official build farm a little bit mad if a release tag's contents changed underneath it.

jspricke:
>@jspricke this pull request has been open for a while and doesn't appear to have rotted at all. Are you still interested in seeing it merged?

Yes, sure :).

>* If we'd favor implementing a package revision option that is not distribution specific. But as each distribution has variation in the concept perhaps distro specific is best...

There is `--rpm-inc` already as well.

>* Do we need to put any protections in to prevent maintainers from re-blooming an earlier debinc/pkgrel? While we can't stop folks from doing silly things on their own projects it would drive the official build farm a little bit mad if a release tag's contents changed underneath it.

That should be handled by the repository management later. Also it was 
always 0 up to now ;). The official build farms adds a time stamp, btw.

PR #432, opened 2017-08-23
Handle utf-8 in more cases.

nuclearsandwich:
encodings

https://github.com/ros-infrastructure/bloom/commit/1f164c156bf1e3b8b4ae1ea053a16c74ade066c1 is focused on git-bloom-import-upstream and https://github.com/ros-infrastructure/bloom/commit/442b4f8a2f49f969d10feaae18182529f1cae31e contains fixes for the Debian generator. Both are tested with python versions
- 2.7.12
- 3.5.2

The test source repository was https://github.com/eProsima/Fast-RTPS and release repository https://github.com/ros2-gbp/fastrtps-release.

If @wjwwood or @mikaelarguedas knows of any ROS kinetic or lunar packages that have utf-8 characters in the upstream or patched in package.xml template I would love to test this explicitly against them too.

I'm pretty sure the way I'm handling some of the py2/3 encoding cases could be improved upon and standardized throughout bloom, but since search-able documentation and Stack Overflow posts aren't in complete agreement about the best way to keep strings compatible between py2 and py3 I may as well just do _something_ and then iterate on it with input from the ROS team. Feel free to make suggestions about alternative implementations.

### DIFF
--- a/bloom/commands/git/import_upstream.py
+++ b/bloom/commands/git/import_upstream.py
@@ -35,6 +35,7 @@ from __future__ import print_function
 import argparse
 import os
 import shutil
+import sys
 import tarfile
 import tempfile
 
@@ -207,10 +208,10 @@ def handle_tree(tree, directory, root_path, version):
             with open(rel_path, 'wb') as f:
                 # Python 2 will treat this as an ascii string but
                 # Python 3 will not re-decode a utf-8 string.
-                try:
-                    file_data = file_data.encode('utf-8')
-                except UnicodeDecodeError:
+                if sys.version_info.major == 2:
                     file_data = file_data.decode('utf-8').encode('utf-8')
+                else:
+                    file_data = file_data.encode('utf-8')
                 f.write(file_data)
             # Add it with git
             execute_command('git add {0}'.format(rel_path), shell=True)
--- a/bloom/commands/git/import_upstream.py
+++ b/bloom/commands/git/import_upstream.py
@@ -38,7 +38,6 @@ import shutil
 import tarfile
 import tempfile
 
-from io import open
 from pkg_resources import parse_version
 
 try:
--- a/bloom/commands/git/import_upstream.py
+++ b/bloom/commands/git/import_upstream.py
@@ -38,6 +38,7 @@ import shutil
 import tarfile
 import tempfile
 
+from io import open
 from pkg_resources import parse_version
 
 try:
@@ -205,6 +206,12 @@ def handle_tree(tree, directory, root_path, version):
                 file_data = show(BLOOM_CONFIG_BRANCH, os.path.join(root_path, rel_path))
             # Write file
             with open(rel_path, 'wb') as f:
+                # Python 2 will treat this as an ascii string but
+                # Python 3 will not re-decode a utf-8 string.
+                try:
+                    file_data = file_data.encode('utf-8')
+                except UnicodeDecodeError:
+                    file_data = file_data.decode('utf-8').encode('utf-8')
                 f.write(file_data)
             # Add it with git
             execute_command('git add {0}'.format(rel_path), shell=True)
--- a/bloom/generators/debian/generator.py
+++ b/bloom/generators/debian/generator.py
@@ -34,6 +34,7 @@ from __future__ import print_function
 
 import collections
 import datetime
+import io
 import json
 import os
 import pkg_resources
@@ -143,8 +144,11 @@ def __place_template_folder(group, src, dst, gbp=False):
             if os.path.exists(template_dst):
                 debug("Removing existing file '{0}'".format(template_dst))
                 os.remove(template_dst)
-            with open(template_dst, 'w') as f:
+            with io.open(template_dst, 'w', encoding='utf-8') as f:
                 if not isinstance(template, str):
+                    template = template.decode('utf-8')
+                # Python 2 API needs a `unicode` not a utf-8 string.
+                elif sys.version_info.major == 2:
                     template = template.decode('utf-8')
                 f.write(template)
             shutil.copystat(template_abs_path, template_dst)
@@ -434,7 +438,9 @@ def __process_template_folder(path, subs):
             os.path.relpath(template_path)))
         result = em.expand(template, **subs)
         # Write the result
-        with open(template_path, 'w') as f:
+        with io.open(template_path, 'w', encoding='utf-8') as f:
+            if sys.version_info.major == 2:
+                result = result.decode('utf-8')
             f.write(result)
         # Copy the permissions
         shutil.copymode(item, template_path)


bloom/commands/git/import_upstream.py
```diff
@@ -205,6 +206,12 @@ def handle_tree(tree, directory, root_path, version):
                 file_data = show(BLOOM_CONFIG_BRANCH, os.path.join(root_path, rel_path))
             # Write file
             with open(rel_path, 'wb') as f:
+                # Python 2 will treat this as an ascii string but
+                # Python 3 will not re-decode a utf-8 string.
+                try:
+                    file_data = file_data.encode('utf-8')
+                except UnicodeDecodeError:
+                    file_data = file_data.decode('utf-8').encode('utf-8')
```
nuclearsandwich: decoding just to re-encode seems less than ideal but maybe it's fine. Is there another way to indicate to python2 when a str is already unicode. With out the explicit `decode('utf-8')` before re-encoding it assumes the string is ascii and fails.

bloom/generators/debian/generator.py
```diff
@@ -34,6 +34,7 @@
 
 import collections
 import datetime
+import io
```
nuclearsandwich: I had initially imported `io.open` directly into the local module but that caused every other instance of `open` to want `unicode`s instead of strings and I decided it would be easier just to use io.open where I explicitly needed it to handle utf-8 input. This could be a footgun but it works for now.

bloom/commands/git/import_upstream.py
wjwwood: Does this actually do anything? I guess it will still throw on a invalid character for decoding as utf-8. Would it be better to do an `if python3 then ...` rather than doing a try-except?

bloom/commands/git/import_upstream.py
```diff
@@ -38,6 +38,7 @@
 import tarfile
 import tempfile
 
+from io import open
```
mikaelarguedas: is this import necessary if we don't need to pass an ecoding to `open`?

bloom/commands/git/import_upstream.py
wjwwood: Sorry started the review before I saw this. I don't if it is necessary, but if it works for your cases I don't think it matters. This isn't a performance issue or anything like that.

mikaelarguedas:
I didnt see many names containing unicode in Lunar so far. I remember [Michael](https://github.com/ros-planning/moveit/blob/7d36403187dffbad7a747e65b98dd266cc0b9da9/moveit/package.xml#L8) from the MoveIt!
 maintainers.

You can scan the maintainer colums on the [status page](http://repositories.ros.org/status_page/ros_kinetic_default.html) to find more variety of unicode characters

bloom/commands/git/import_upstream.py
mikaelarguedas: Sorry I commented before seeing [this comment ](https://github.com/ros-infrastructure/bloom/pull/432#discussion_r134864835), I guess in that case we can do the same here so that "open" always refers to the builtin `open` and `io.open` to the fancier version.

bloom/commands/git/import_upstream.py
nuclearsandwich: A minimal representation of the problem in py2. Note that if you enter the string as `"Ragnarök"` it works fine because something about the interactive python interpreter recognizes it as a utf-8 string already.

```python
"Ragnar\xc3\xb6k".encode('utf-8')
# Traceback (most recent call last):
# File "<stdin>", line 1, in <module>
# UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 6: ordinal not in range(128)
#
# 
"Ragnar\xc3\xb6k".decode('utf-8').encode('utf-8')
# ==> 'Ragnar\xc3\xb6k
```

nuclearsandwich: > Does this actually do anything? I guess it will still throw on a invalid character for decoding as utf-8. Would it be better to do an if python3 then ... rather than doing a try-except?

Sadly there's enough going on in the UI here that I'm not sure which line this is questioning. Can you drop a line number or quote some code? Thanks!

wjwwood: Line 214, `file_data = file_data.decode('utf-8').encode('utf-8')`

wjwwood: But generally the whole 211-214 section (w.r.t. possibly replacing it with an if-else)

bloom/commands/git/import_upstream.py
nuclearsandwich: > is this import necessary if we don't need to pass an ecoding to open?

I know that I only worked on this a couple hours ago but I've already forgotten. If it is needed it will be because of the different types expected by the `open` builtin versus `io.open`. I'll revert this change and retest to be sure.



> I guess in that case we can do the same here so that "open" always refers to the builtin open and io.open to the fancier version.

This is a trade-off between having one consistent API for `open` and only using `io.open` when the python2 builtin doesn't work. `io.open` has the same API in Python 2 and Python 3 and is encoding aware.

Using it everywhere would cut down on some mental overhead and would future-proof code against utf-8 characters at the cost of requiring utf-8 handling everywhere. If we only use `io.open` where we know that we need it we avoid that overhead but risk future errors when utf-8 pops up where there was only ascii previously.

That being said, future errors are speculative and mental overhead is a surety so it's really a question of which is more work for the programmer: Sometimes using `io.open` or always using `io.open`?

mikaelarguedas: I don't have authority here so I'll leave it up to @wjwwood to decide. I'd personally lean toward using `io.open` everywhere to avoid having to think about it. To keep it explicit, I'd call explicitly `io.open` so that we dont need to check the beginning of the file to know if it has been imported from `io` or not

wjwwood: Doing `import io` and using `io.open()` everywhere is probably slightly better as to not overwrite a built-in, but I personally don't mind either way.

bloom/commands/git/import_upstream.py
nuclearsandwich: I'm happy replacing this with an explicit python version check. I'd noticed other places, particularly in imports where exception catching was used for 2/3 compatibility and was trying to keep consistent where I could, but I like the explicit version checks better.

The explicit decode before re-encoding was the most plain way I could find to make file_data into a utf-8 string without changing the behavior of execute_command.

bloom/commands/git/import_upstream.py
nuclearsandwich: I can switch these up to use `io.open`. The big downside to using io.open everywhere is that in python2 the IO object wants `unicode`s not strings whereas it's perfectly happy with strings in python3 and even if we wanted to just throw in utf-8 bytes anyways the `unicode()` cast doesn't exist for py3 and nor is decode() defined for strings.

I think this means that unless we resolve this a completely different way, every time we read or write a file with `io.open` we'll need a python version check.

mikaelarguedas: hmmm yeah that's unfortunate, Python2 and encoding is always a pain :(
Well I guess, my recommendation would then be: if its confirmed that we need this import here let's call out `io.open` explicitely otherwise let's ditch the import in this file.
And for the other files lets not override the default to avoid unwanted side effects (having to convert every string to unicode) and just use io.open where it's necessary.

Thanks for the detailed explanation!

nuclearsandwich:
I think this is ready for final review. I know I've got the green checks but I've made some changes in response to conversation which should get a quick look before merging.

bloom/commands/git/import_upstream.py
```diff
@@ -205,6 +206,12 @@ def handle_tree(tree, directory, root_path, version):
                 file_data = show(BLOOM_CONFIG_BRANCH, os.path.join(root_path, rel_path))
             # Write file
             with open(rel_path, 'wb') as f:
+                # Python 2 will treat this as an ascii string but
+                # Python 3 will not re-decode a utf-8 string.
+                if sys.version_info.major == 2:
+                    file_data = file_data.decode('utf-8').encode('utf-8')
+                else:
+                    file_data = file_data.encode('utf-8')
```
wjwwood: I think this looks better. If in the future using a try-except is easier just add a comment that this is expected to fail on Python2/3 and not for some other reason.

PR #434, opened 2017-08-31
Add support for ROS 2 build types.

nuclearsandwich:
This PR is directly based on the work in https://github.com/ros2/bloom.

It adds support for the ROS 2 build types and needed features by replacing explicit template contents with template variables that mostly come from the `set_build_type_options` function.

This is the minimal changeset required to support packaging ROS 2 for Beta 3 and is meant as a foil PR to https://github.com/ros-infrastructure/bloom/pull/435 which contains a reorganization of the debian templates.

There is some work still outstanding if this candidate is to be merged: Critically: I haven't yet re-added support for catkin packages. I also need to add some tests that will cover the ROS 2 packages.

### DIFF
--- a/bloom/generators/debian/generator.py
+++ b/bloom/generators/debian/generator.py
@@ -543,8 +543,9 @@ def set_debhelper_options(data, package):
             'toplevel': '--buildsystem=pybuild --with python3',
             'depends': '${python3:Depends}',
             'exportvars': [('PYBUILD_INSTALL_ARGS',
-                           '--prefix="{0}" --install-lib="{0}/lib/{{interpreter}}/site-packages"'
-                            .format(data['InstallationPrefix']))],
+                           '--prefix="{0}" --install-lib="$base/lib/{{interpreter}}/site-packages" '
+                           '--install-scripts="$base/bin"'
+                           .format(data['InstallationPrefix']))],
         },
         'cmake': {
             'toplevel': '--buildsystem=cmake',
--- a/bloom/generators/debian/generator.py
+++ b/bloom/generators/debian/generator.py
@@ -513,20 +513,22 @@ def debianize_string(value):
 def sanitize_package_name(name):
     return name.replace('_', '-')
 
+
 def set_debhelper_options(data, package):
     debhelper_options = {
-            'ament_cmake': {
-                'toplevel': '--buildsystem=cmake',
-                'autoconfigure': '-- -DCMAKE_INSTALL_PREFIX="{0}" -DAMENT_PREFIX_PATH="{0}"'.format(data['InstallationPrefix'])
-                },
-            'ament_python': {
-                'toplevel': '--buildsystem=pybuild --with python3',
-                },
-            'cmake': {
-                'toplevel': '--buildsystem=cmake',
-                'autoconfigure': '-- -DCMAKE_INSTALL_PREFIX="{0}"'.format(data['InstallationPrefix'])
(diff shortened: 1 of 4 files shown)


bloom/generators/debian/generator.py
```diff
@@ -511,6 +516,77 @@ def sanitize_package_name(name):
...
+    if build_type == 'ament_cmake':
+        pass
+    elif build_type == 'ament_python':
+        # XXX inject build type specific dependencies.  For now this needs to be done
+        # after resolving rosdep keys as these are Debian platform-specific dependencies.
+        data['BuildDepends'] = list(set([
+                                        'dh-python',
+                                        'python3-all (>= 3.5)',
+                                        'python3-setuptools'
+                                        ] + data['BuildDepends']))
```
wjwwood: nitpick, but I would have formatted this so (basically I always avoid indenting more than once if possible):

```python
data['BuildDepends'] = list(set([
    'dh-python',
    'python3-all (>= 3.5)',
    'python3-setuptools'
] + data['BuildDepends']))
```

bloom/generators/debian/generator.py
```diff
@@ -511,6 +516,77 @@ def sanitize_package_name(name):
...
+        # Set `--install-scripts` unless it is specified in setup.cfg.
+        pass_install_scripts = True
+        package_path = os.path.abspath(os.path.dirname(package.filename))
+        setup_cfg_path = os.path.join(package_path, "setup.cfg")
+        if os.path.isfile(setup_cfg_path):
+            from configparser import SafeConfigParser
```
wjwwood: I usually try to avoid local imports unless there is a performance or dependency minimization reason. By having them all at the top of the file it is easier to see, at a glance, what's being used in the file. Again, this is a nitpick not a requested change.

bloom/generators/debian/generator.py
```diff
@@ -511,6 +516,77 @@ def sanitize_package_name(name):
...
+                pass_install_scripts = False
+        if pass_install_scripts:
+            data['exportvars']['PYBUILD_INSTALL_ARGS'] += ' --install-scripts="\$$base/bin" '
+    elif build_type == 'cmake':
+        pass
+    else:
+        # Even though we check supported build_types in debhelper options,
+        # preserve this so that forgetting to add a new build type to both the debhelper
+        # options and this conditional block will raise an error.
+        error('The build type `{build_type}`is not supported by this version of bloom.'
```
wjwwood: nitpick, the standard style we use when quoting something in a message like this is to use `"` to capture the string, and `'` to quote the detail you want quoted. I'm not sure why this is prefered, but it's the common style. For example:

- https://github.com/ros-infrastructure/bloom/blob/master/bloom/config.py#L159
- https://github.com/ros-infrastructure/bloom/blob/master/bloom/git.py#L279
- https://github.com/ros-infrastructure/rosdistro/blob/master/src/rosdistro/source_build_file.py#L44

But of course there are always counterexamples:

- https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/python_setup.py#L115

The important thing is that you're quoting it 👍, but if you want you can update it to be more consistent with the other code we have.

nuclearsandwich:
The code feedback for both this PR and #435 is really valuable, but before making more changes I think we need to decide which approach to stick with.

nuclearsandwich:
Closing in favor of #435

PR #438, opened 2017-10-19
use https rather than http to query pypi

wjwwood:
Otherwise I get:

> Error fetching latest version: <ProtocolError for pypi.python.org/pypi: 403 Must access using HTTPS instead of HTTP>

### DIFF
--- a/bloom/commands/update.py
+++ b/bloom/commands/update.py
@@ -106,7 +106,7 @@ def fetch_update(user_bloom):
     if os.path.exists(user_bloom):
         return
     open(user_bloom, 'w').close()  # Touch the file
-    pypi = ServerProxy('http://pypi.python.org/pypi')
+    pypi = ServerProxy('https://pypi.python.org/pypi')
     newest_version = pypi.package_releases('bloom')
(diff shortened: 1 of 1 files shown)


wjwwood:
Also, pypi currently reports that `0.5.26` is the latest (and only) version of bloom via this mechanism. Maybe it's just caching of some sort, but it's worth checking on again (can check with `bloom-update`).

nuclearsandwich:
I tested this again today and the xmlrpc API was still reporting 0.5.26 as latest. I tried using the JSON API and it worked just fine so I went ahead and switched us to that. The bloom-update command was tested by setting the locally installed version to 0.0.99 and running in various py2 and py3 virtualenvs.

wjwwood:
lgtm (can't approve my own pr)

tfoote:
Note possibly related to older versions being reported there's general infrastructure changes going on at pypi. pypi.python.org is migrating to pypi.org: https://packaging.python.org/guides/migrating-to-pypi-org/

nuclearsandwich:
Huh, I found pypi.org while I was testing this but was chased away by the scary red pre-production banner.

Issue #448, opened 2017-11-27
Failed to open pull request: AttributeError - 'GithubException' object has no attribute 'resp'


130s:
```
$ bloom-release --rosdistro kinetic --track kinetic geometric_shapes -p
:
       url: https://github.com/ros-planning/geometric_shapes.git

Looks like bloom doesn't have an oauth token for you yet.
Therefore bloom will require your GitHub username and password just this once.
With your GitHub username and password bloom will create an oauth token on your behalf.
The token will be stored in `~/.config/bloom`.
You can delete the token from that file to have a new token generated.
Guard this token like a password, because it allows someone/something to act on your behalf.
If you need to unauthorize it, remove it from the 'Applications' menu in your GitHub account page.

Would you like to create an OAuth token now [Y/n]? 
GitHub username [n130s]: 130s
GitHub password (never stored): 
HTTP Error 401: Unauthorized (https://api.github.com/authorizations): None

Failed to open pull request: AttributeError - 'GithubException' object has no attribute 'resp'

$ ll ~/.config/bloom
ls: cannot access '/home/n130s/.config/bloom': No such file or directory

$ tail -n 25 ~/.bloom_logs/bloom-release_2017-11-26_16.35.06.885019.log
[info]      source:
[info]        type: git
[info]        url: https://github.com/ros-planning/geometric_shapes.git
[info] 
[warning] Looks like bloom doesn't have an oauth token for you yet.
[warning] Therefore bloom will require your GitHub username and password just this once.
[warning] With your GitHub username and password bloom will create an oauth token on your behalf.
[warning] The token will be stored in `~/.config/bloom`.
[warning] You can delete the token from that file to have a new token generated.
[warning] Guard this token like a password, because it allows someone/something to act on your behalf.
[warning] If you need to unauthorize it, remove it from the 'Applications' menu in your GitHub account page.
[info] 
[error] HTTP Error 401: Unauthorized (https://api.github.com/authorizations): None
[info] 
[debug] Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/bloom/commands/release.py", line 1239, in perform_release
    track, repository, distro, interactive, override_release_repository_url
  File "/usr/lib/python2.7/dist-packages/bloom/commands/release.py", line 768, in open_pull_request
    gh = get_github_interface()
  File "/usr/lib/python2.7/dist-packages/bloom/commands/release.py", line 692, in get_github_interface
[...]

PR #451, opened 2017-12-13
GitHub PR auth prompt fixes

nuclearsandwich:
This is a proposed fix for #448.

A detailed outline of the code paths is present in that issue.
The error interface appears to have changed since the original code was written.
This adds code to handle the new error interface while preserving the old code.
It's possible this old code is no longer used and can be removed but I wouldn't know how to tell since #448 surfaces for me using both Python 2 and Python 3.


nuclearsandwich:
@wjwwood this could still do with a review when you have a chance.

Issue #181, opened 2013-07-09
bloom-release track and rosdistro arguments are redundant


ahendrix:
The `--track` and `--rosdistro` arguments to bloom-release are largely redundant. I propose two possible solutions:
- If `--rosdistro` is specified and `--track` is not, assume they are the same
  OR
- If `--rosdistro` is specified and `--track` is not, go to the appropriate distro file, find the release repository, pull down the track config from the release repository, and find the track that matches the specified rosdistro. Prompt the user if there is more than one track that matches.

DLu:
Closed by #459

trainman419:
Awesome! Thank you @DLu !

PR #459, opened 2018-02-05
Optional track

DLu:
One small change that has always bugged me: This PR makes the track command line argument optional, and if it is not set, it sets it to the rosdistro.


nuclearsandwich:
I added the default value to the helpline. Feel free to make additional edits to the language if you prefer.

The change looks good to me but I've asked @wjwwood for thoughts as he's got more history than I do.

wjwwood:
https://github.com/ros-infrastructure/bloom/issues/181

DLu:
All set?

PR #460, opened 2018-02-05
RELEASE REPO BASE env var

DLu:
* Enables the use of a new environment variable `BLOOM_RELEASE_REPO_BASE`. 
   * This can be used instead of manually inputting the release repo url. 
   * If the environment var exists, and `BLOOM_RELEASE_REPO_BASE` + repository + '-release.git' exists as a repository, then it will use that as the release repo.
   * For example, all my release repos are over at `https://github.com/wu-robotics/` so if I set that as my environment variable, I can create a `bloom-example-release` repo there, and then if I try to `bloom-release` a package called `bloom-example`, it will automagically use that as my release repo.

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -44,6 +44,7 @@ import json
 import os
 import pkg_resources
 import platform
+import requests
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
I know bloom's documentation isn't in the best of shape, but I would love to see this new feature documented somewhere so I can add documentation for more of the environment variables used by Bloom.

bloom/commands/release.py
```diff
@@ -296,6 +297,16 @@ def validate_github_url(url, url_type):
     return True
 
 
+def infer_release_repo_from_env(repository):
```
nuclearsandwich: Can you please add a docstring to this function describing the environment variable's behavior.

DLu:
I would be willing to add a test or some documentation, but I'm unclear where it would go best. I'm not familiar with the test structures in this package, and it seems like the documentation should just go on the wiki.

nuclearsandwich:
Thanks for the docstring DLu. I'll admit that I'm also getting familiar with the test structures so I don't have much concrete advice for you on that front.

I'm going to merge this so we can add support in the next release.

Issue #217, opened 2014-01-22
support upstream git repositories with submodules


wjwwood:
See: https://github.com/ros-infrastructure/bloom/issues/202

This will require a patch in `vcstools`.

asmodehn:
I am using bloom-release 0.5.20 and I still experience this problem.
I d like to solve it asap since I use a few repositories with submodules ( to be able to embed extra uptodate python-only packages into a ros package. )
Please , let me know the best way I can help, thanks.

tfoote:
There's a workaround mentioned in #202 and you could probably locally make the changes in vcstools/vcstools#88 or vcstools/vcstools#95 

To help solve this properly you could work on solving the outstanding problems in those tickets so it can be released into `vcstools` which `bloom` builds upon.

asmodehn:
See https://github.com/vcstools/vcstools/pull/108 for a fix for https://github.com/vcstools/vcstools/pull/95
Thanks for letting me know if anything else is needed, there or somewhere else.
I might have a tiny bit of time to spend on this, since it is critical for me right now.

asmodehn:
So if anyone needs this feature urgently, I recently tested with : 
- trusty
- indigo
- bloom 0.5.20
- vcstools https://github.com/vcstools/vcstools/commit/dbe026c2c956d3fde808185e525a027a3b658828 from https://github.com/vcstools/vcstools/pull/108
- git > v1.8.3
- buildbot-ros with this change https://github.com/yujinrobot/buildbot-ros/commit/d3ccbd8b2059d747c2c30197e79aebfdfd093ade

And I could release a package from a repository containing a submodule : https://github.com/asmodehn/celeros/tree/indigo-devel/deps 
The (private) release repo contains the submodule,and I was able to build on buildbot-ros.
It's currently sitting in a private release repository, but I ll release it on public rosdistro when all the submodules issues are solved.

k-okada:
are we ready for this? if so, we can solve https://github.com/jsk-ros-pkg/jsk_recognition/issues/1817 issue

wjwwood:
Have you tried releasing a package with this pr of bloom? It would be good to see that it works before merging it with master and releasing it for everyone that uses bloom.

wjwwood:
Ah, this isn't a pr. @asmodehn are you saying that bloom works as-is? The upstream vcstools pr's (or their replacements) have been merged and I think released as well.

k-okada:
[...]

PR #461, opened 2018-02-11
don't error if git repo has submodules

flixr:
although it seems this error was not even triggered before...
In any case releasing an upstream repo with submodules works with
https://github.com/vcstools/vcstools/pull/130

With the vcstools fix, we should be able to close https://github.com/ros-infrastructure/bloom/issues/217
At least it worked for me...

### DIFF
--- a/bloom/commands/export_upstream.py
+++ b/bloom/commands/export_upstream.py
@@ -51,7 +51,6 @@ from bloom.logging import warning
 
 from bloom.git import branch_exists
 from bloom.git import get_root
-from bloom.git import has_submodules
 from bloom.git import tag_exists
 
 from bloom.util import add_global_arguments
--- a/bloom/commands/export_upstream.py
+++ b/bloom/commands/export_upstream.py
@@ -121,14 +121,6 @@ def export_upstream(uri, tag, vcs_type, output_dir, show_uri, name):
                 error("Failed to clone repository at '{0}'".format(uri) +
                       (" to reference '{0}'.".format(tag) if tag else '.'),
                       exit=True)
-        if get_root() is not None and has_submodules(upstream_repo.get_path()):
-            error("""\
-bloom does not support exporting git repositories with submodules, see:
-
-- https://github.com/ros-infrastructure/bloom/issues/202
-- https://github.com/ros-infrastructure/bloom/issues/217
-- https://github.com/vcstools/vcstools/issues/84
-""", exit=True)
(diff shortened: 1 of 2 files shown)


flixr:
Ah, and for completeness: for the prerelease job this was also needed: https://github.com/ros-infrastructure/ros_buildfarm/pull/515

We'll see if there is anything else missing in the buildfarm when the rc_visard (https://github.com/ros/rosdistro/pull/16954) job gets picked up, but hopefully that was all now... ;-)

nuclearsandwich:
@flixr thanks for the contributions. 

> although it seems this error was not even triggered before...

Can you explain what you mean here. Did you encounter this error when working on your release? Or is this a cleanup pass since this error doesn't appear necessary anymore?

Thanks.

flixr:
> although it seems this error was not even triggered before...
> 
> Can you explain what you mean here. Did you encounter this error when working on your release? Or is this a cleanup pass since this error doesn't appear necessary anymore?

In my case this error was not triggered (although it probably should have been). I think there a path should have been passed to `get_root()`...
In any case this error is not needed anymore.

flixr:
Yes, I did a bloom release for https://github.com/roboception/rc_visard (which has a few nested submodules) into https://github.com/roboception/rc_visard-release
Looks OK to me.
See PR https://github.com/ros/rosdistro/pull/16954

flixr:
Just to be clear, I also needed https://github.com/vcstools/vcstools/pull/130 for it to work with the relative submodule URLs specified in https://github.com/roboception/rc_visard_ros/blob/master/.gitmodules
It was probably already working for absolute URLs.

nuclearsandwich:
> In my case this error was not triggered (although it probably should have been).

Thanks. I am trying to determine whether we need fix or remove our `has_submodules` function and it looks like we do.

nuclearsandwich:
@wjwwood I think it's probably best to just remove the has_submodules function as it is not used outside of this error message. Would you agree? We can always fetch it from the git history if it proves needed in the future.

wjwwood:
@nuclearsandwich yes, that sounds fine to me.

dirk-thomas:
@nuclearsandwich @wjwwood Friendly ping.

nuclearsandwich:
I want to remove the has_submodules function entirely (which @flixr is welcome to add to this PR) and merge this for bloom's 0.6.3 release later this week.

flixr:
done...

nuclearsandwich:
:confetti_ball: perfect. Thanks!

PR #466, opened 2018-03-12
Test on Python 3.4 and 3.6 because 3.3 is EOL

cclauss:
Current versions of Python: https://devguide.python.org/#branchstatus

Python 3.3 reached its end of life last year.


wjwwood:
Thanks!

PR #470, opened 2018-03-31
Generate debian/copyright file from license tag

otamachan:
Hi,
this is implementation in progress for https://github.com/ros-infrastructure/rep/pull/157

See also https://github.com/ros-infrastructure/catkin_pkg/pull/202

1. ``copyright`` file is generated just by concatenating license files.
2. Even when the format version is less than 3, an empty copyright file will be generated.

Thanks.


bloom/generators/debian/generator.py
```diff
@@ -417,6 +417,19 @@ def generate_substitutions_from_package(
...
+        if hasattr(l, 'file') and l.file is not None:
+            license_file = os.path.join(os.path.dirname(package.filename), l.file)
+            if not os.path.exists(license_file):
+                error("License file '{}' is not found.".
+                      format(license_file), exit=True)
+            license_text = open(license_file, 'r').read()
+            if not license_text.endswith('\n'):
+                license_text += '\n'
+            licenses.append(license_text)
+    data['Copyright'] = '\n'.join(licenses)
```
nuclearsandwich: Do you have any in-the-wild example packages that you've tested this on? Depending on the contents of the individual license files a single newline between them may not be very readable.

I'm thinking of a case like
```
the final line of one license file.
Copyright 2015-2018 The authors of Another Project In The Package

The start of this second license text...
```

nuclearsandwich:
> Even when the format version is less than 3, an empty copyright file will be generated.

Is generating an empty copyright file the right thing to do? I know that lintian wants a copyright file always, but I'm not sure that an empty copyright file is better for our packages than no copyright file if the package's license is found in the package.xml and we cannot determine where the license file is.

In essence, my question is: Should this functionality be restricted to format 3 packages only to avoid confusing the license issue with stray files?

Aside from that, the changes look good on initial review. I appreciate the time spent improving the test fixture generation.

otamachan:
The generation of an empty copyright file when the format version is less than 3 is just because it's easier to implement. I will try to avoid it.

otamachan:
@nuclearsandwich I'm sorry for my late response, but I think it's ready now. Could you review this again? Thanks in advance.

130s:
@nuclearsandwich wrote:
> Is generating an empty copyright file the right thing to do? I know that lintian wants a copyright file always, but I'm not sure that an empty copyright file is better for our packages than no copyright file if the package's license is found in the package.xml and we cannot determine where the license file is.
> 
> In essence, my question is: Should this functionality be restricted to format 3 packages only to avoid confusing the license issue with stray files?

+1

One usecase out of potentially many for this feature is to build an application to collect license files and do whatever processing (e.g. https://github.com/ros-infrastructure/rospkg/pull/143). Populating empty license files could that kind of downstream processing more difficult and cumbersome.

nuclearsandwich:
> @nuclearsandwich I'm sorry for my late response, but I think it's ready now. Could you review this again? Thanks in advance.

Quick update to let you know that this PR isn't forgotten! I'm really busy with the upcoming ROS 2 release these days. Once it is released and the dust has settled I look forward to reviewing this PR again. Thanks for your patience.

bloom/generators/debian/generator.py
```diff
@@ -417,6 +417,20 @@ def generate_substitutions_from_package(
     data['debhelper_version'] = 7 if os_version in ['oneiric'] else 9
     # Summarize dependencies
     summarize_dependency_mapping(data, depends, build_depends, resolved_deps)
+    # Copyright
+    licenses = []
+    separator = '\n' + '=' * 80 + '\n\n'
```
nuclearsandwich: Why this choice for the license separator?

bloom/generators/debian/generator.py
```diff
@@ -471,6 +485,9 @@ def __process_template_folder(path, subs):
             os.path.relpath(item),
             os.path.relpath(template_path)))
         result = em.expand(template, **subs)
+        # Skip if content is empty (for copyright)
+        if len(result) == 0:
```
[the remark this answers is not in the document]
otamachan: OK, I will scope this logic only to copyright file.

otamachan:
> Did you consider the Debian machine readable copyright format at all? I don't think we necessarily have sufficient information from the package.xml for the copyright info or multiple licenses.

Yes, I know that format, but I couldn't find the proper way to fill ``Files:`` field as you said.

> I also haven't looked at any similar facilities in the RPM templates.

Sorry, I'm not familiar with RPMs. It seems that they use ``%license`` macro for this pupose, but I cannot test it at all.
* http://rpm-guide.readthedocs.io/en/latest/rpm-guide.html#working-with-spec-files
* https://fedoraproject.org/wiki/Packaging:LicensingGuidelines?rd=Packaging/LicensingGuidelines

bloom/generators/debian/generator.py
otamachan: Because I just found some packages use this as a separator.
But now I found the "\n----*\n\n" is used more than "\n===*\n\n" in my computer. Hmm

```bash
/usr/share/doc⟫ grep -PHzol "\n\n======*\n\n" */copyright | wc -l
24
/usr/share/doc⟫ grep -PHzol "\n\n------*\n\n" */copyright | wc -l
55
```

Do you have better idea?

nuclearsandwich:
> Sorry, I'm not familiar with RPMs. It seems that they use %license macro for this pupose, but I cannot test it at all.

Thanks for the links. I don't think RPM support is a requirement to merge this PR.  My main goal in asking was to make sure we didn't back ourselves into a corner with the Debian format if we decided to try and use it and I don't think we do.

> Yes, I know that format, but I couldn't find the proper way to fill Files: field as you said.

I conferred privately with @j-rivero about the Debian copyright format and I will summarize that conversation but he is welcome to contradict me. Both of us think that moving packages bloom generates closer to compliance with Debian guidelines is a good thing. His suggestion was to generate Debian compatible data and omit fields that we don't have necessary knowledge to supply.

But I think the best way for bloom to support the Debian format at the moment is to encourage upstream sources to adopt Debian's machine readable format if they want to use it. That prevents bloom from trying to do any mangling or manipulation of the upstream license file which may unintentionally change its meaning which in my opinion is the worst possible offense where license/copyright communication is concerned.

bloom/generators/debian/generator.py
nuclearsandwich: Given that there is no overriding convention, I have no preference. I think either convention is fine and it can be left as-is.

nuclearsandwich:
I was just doing some local testing before merging this. There is one small issue which is that for repositories without a copyright file the copyright.em template file is being copied into the expanded directory.

bloom/generators/debian/generator.py
```diff
@@ -471,6 +485,10 @@ def __process_template_folder(path, subs):
             os.path.relpath(item),
             os.path.relpath(template_path)))
         result = em.expand(template, **subs)
+        # Skip the result is empty
```
nuclearsandwich: Looks like a small word omission typo. Perhaps change it to "Skip if the result is empty" or "Don't write an empty file"?

bloom/generators/debian/generator.py
```diff
@@ -471,6 +485,10 @@ def __process_template_folder(path, subs):
...
+        if len(result) == 0 and \
+           os.path.basename(template_path) in ['copyright']:
+            continue
```
nuclearsandwich: Before continuing here we need to mark the template file as processed so it will not b copied verbatim to the processed directory. Adding the line `processed_items.append(item)` should do the trick.

otamachan:
Thank you for your review!
I just rebased the branch and fixed the code.

PR #472, opened 2018-05-11
Upstream Tag

DLu:
If `:{ask}` is specified for the release tag, the upstream tag is not written into the release repository, making it quite difficult to figure out what tag was used. This PR ensures that it is written into `tracks.yaml`.

### DIFF
--- a/bloom/commands/git/release.py
+++ b/bloom/commands/git/release.py
@@ -304,6 +304,9 @@ def execute_track(track, track_dict, release_inc, pretend=True, debug=False, fas
         tracks_dict = get_tracks_dict_raw()
         tracks_dict['tracks'][track]['release_inc'] = settings['release_inc']
         tracks_dict['tracks'][track]['last_version'] = settings['version']
+        # if release tag is set to ask and a custom value is used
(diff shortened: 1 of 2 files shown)


mikepurvis:
The situation is even worse if you specify some arbitrary git hash as the "tag". This has been a long time coming— strong support here.

DLu:
Who is the right person to review this? @wjwwood ?

nuclearsandwich:
Thanks for the patience. I don't want to bikeshed too much but I wonder if `current_release` would make a better key name. When I read `last_release` I think of release N-1 rather than release N. But if `last_release` is clear enough to others I'm not strongly opposed.

DLu:
I'm open to changing the name. However, the motivation was to mirror [last_version](https://github.com/locusrobotics/bloom/blob/d6ea291936c11d5a024617694bd8fbdb67187d20/bloom/commands/git/release.py#L306)

nuclearsandwich:
> I'm open to changing the name. However, the motivation was to mirror last_version

I think precedent and consistent naming overrides the possibility of an incorrect assumption. :+1:

PR #473, opened 2018-06-07
Revert "Add bouncy to the list of rosdistros to get the workspace shim."

dirk-thomas:
This reverts #462.

For Bouncy we won't use the `ros_workspace` repository but let the `ament_cmake_core` package provide the setup files instead using [this option](https://github.com/ament/ament_cmake/blob/4e46886895a9f97c0ef82c4fedacd13dcb010aac/ament_cmake_core/ament_cmake_environment-extras.cmake#L17-L18).


nuclearsandwich:
This approach is solid because it preserves the behavior for the distros that need it. Actually approving and merging is pending a decision on whether using ament_cmake_core is the desired course of action.

mikaelarguedas:
As the currently released version meets our needs for bouncy, we will delay this PR until the C-turtle release cycle and decide to use tis change or not based on the decision of how we will provide environment scripts to users in C-Turtle

PR #474, opened 2018-06-22
Reference the LICENSE.txt from the stdeb.cfg

tfoote:
stdeb will install it in the debian way when referenced here.


nuclearsandwich:
Looks good to me. The license attributes copyright to solely to Willow Garage and hasn't been  updated to include years after 2013. I think it would be valuable to make any updates to the license that need to be made to maximize the value of this PR.

PR #475, opened 2018-06-22
Inject vendor typesupport packages at build time for ROS 2.

nuclearsandwich:
In order to build packages which support but do not require all rmw providers we need to prevent them from being required by end users but make sure that they're installed on the buildfarm when any message package is built.

To avoid needing to apply these changes manually we've added another ROS 2-specific behavior to bloom.

The new behavior looks at group membership and adds vendor typesupport packages to the build (but not runtime or build export) dependencies of the packages in the rosidl_interface_packages group.

### DIFF
--- a/bloom/generators/rosdebian.py
+++ b/bloom/generators/rosdebian.py
@@ -87,12 +87,29 @@ class RosDebianGenerator(DebianGenerator):
         )
         subs['Package'] = rosify_package_name(subs['Package'], self.rosdistro)
 
-        # XXX Add workspace package to runtime and buildtime dependencies for ROS 2 only.
-        if self.rosdistro in ['r2b2', 'r2b3', 'ardent', 'bouncy'] and \
-                package.name not in ['ament_cmake_core', 'ament_package', 'ros_workspace']:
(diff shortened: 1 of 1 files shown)


bloom/generators/rosdebian.py
```diff
@@ -87,12 +87,27 @@ def fallback_resolver(key, peer_packages, rosdistro=self.rosdistro):
...
+                subs['Depends'].append(workspace_pkg_name)
+
+            # Add vendor typesupport packages to build dependencies for rosidl_interface_packages.
+            if rosdistro in ['bouncy'] and 'rosidl_interface_packages' in package.member_of_groups:
+                ROS2_VENDOR_TYPESUPPORT_PACKAGES = [
+                        'rosidl_typesupport_connext_c',
+                        'rosidl_typesupport_connext_cpp',
+                        'rosidl_typesupport_opensplice_c',
+                        'rosidl_typesupport_opensplice_cpp',
+                        ]
```
wjwwood: Nitpick: The indentation here is strange, I would have expected only one set of four spaces before the items and the closing bracket to align with the start of the variable name.

PR #477, opened 2018-06-28
Add rmw implementations to vendor typesupport build dependencies.

nuclearsandwich:
The additional rmw implementation packages need to be present at build time for vendor
typesupport to be configured properly.

Follow up to #475.

### DIFF
--- a/bloom/generators/rosdebian.py
+++ b/bloom/generators/rosdebian.py
@@ -98,10 +98,15 @@ class RosDebianGenerator(DebianGenerator):
                 subs['BuildDepends'].append(workspace_pkg_name)
                 subs['Depends'].append(workspace_pkg_name)
 
-            # Add vendor typesupport packages to build dependencies for rosidl_interface_packages.
+            # Add packages necessary to build vendor typesupport for rosidl_interface_packages to their
+            # build dependencies.
             if self.rosdistro in ['bouncy'] and \
                     'rosidl_interface_packages' in [p.name for p in package.member_of_groups]:
-                ROS2_VENDOR_TYPESUPPORT_PACKAGES = [
+                ROS2_VENDOR_TYPESUPPORT_DEPENDENCIES = [
+                    'rmw-connext-cpp',
+                    'rmw-fastrtps-cpp',
+                    'rmw-implementation',
+                    'rmw-opensplice-cpp',
                     'rosidl-typesupport-connext-c',
                     'rosidl-typesupport-connext-cpp',
                     'rosidl-typesupport-opensplice-c',
@@ -109,7 +114,7 @@ class RosDebianGenerator(DebianGenerator):
                 ]
 
                 subs['BuildDepends'] += [
-                    rosify_package_name(name, self.rosdistro) for name in ROS2_VENDOR_TYPESUPPORT_PACKAGES]
+                    rosify_package_name(name, self.rosdistro) for name in ROS2_VENDOR_TYPESUPPORT_DEPENDENCIES]
         return subs
 
     def generate_branching_arguments(self, package, branch):


dirk-thomas:
I don't understand why interface packages should depend on the rmw packages? If the already injected typesupport packages need anything else can't they pull it in on their own?

nuclearsandwich:
> If the already injected typesupport packages need anything else can't they pull it in on their own?

The typesupport packages are dependencies of the rmw packages. They can't depend on rmw packages without introducing a cycle.

I don't have a complete understanding of the logic which decides how to build and link interface typesupport. But without the vendor rmw packages present at build time message typesupport links only against rosidl_typesupport_introspection*.so

The rmw packages depend on the typesupport packages so we could remove the direct typesupport packages from the list of injected dependencies. I wouldn't be opposed to that.

dirk-thomas:
The point is that interface packages as well as typesupport packages are intentionally not depending on rmw packages. I don't think that should be changed since that is a conscious architectural decision.

There needs to be very strong and clear evidence that is must be done which I don't see here.

nuclearsandwich:
> There needs to be very strong and clear evidence that is must be done which I don't see here.

The cmake logic in the function `get_used_typesupports` [here](https://github.com/ros2/rosidl_typesupport/blob/45f5a0ad4d9eaa5fc9303be2633a1e306b84026f/rosidl_typesupport_c/cmake/get_used_typesupports.cmake#L26-L29) queries the ament resource index for "rmw_typesupport"[sic] resources. In my local workspace these resources are only supplied by the packages rmw_connext_cpp, rmw_fastrtps_cpp, rmw_implementation, and rmw_opensplice_cpp

```
❯ fd -tf -p '.*share/ament_index/resource_index/rmw_typesupport/.*' install/ 
install/rmw_connext_cpp/share/ament_index/resource_index/rmw_typesupport/rmw_connext_cpp
install/rmw_fastrtps_cpp/share/ament_index/resource_index/rmw_typesupport/rmw_fastrtps_cpp
install/rmw_implementation/share/ament_index/resource_index/rmw_typesupport/rmw_implementation
install/rmw_opensplice_cpp/share/ament_index/resource_index/rmw_typesupport/rmw_opensplice_cpp
```

Since interface packages depend on `rosidl_default_generators` depends on `rosidl_generator_py` depends on `rmw_implementation` which will satisfy its dependency with `rmw_fastrtps_cpp` by default.

--

The logic above is triggered by the cmake macro [rosidl_generate_interfaces](https://github.com/ros2/rcl_interfaces/blob/d209af6a5d3c631c7059469f4f30983a57d8b950/builtin_interfaces/CMakeLists.txt#L18) -> [ament_execute_extensions("rosidl_generate_interfaces")](https://github.com/ros2/rosidl/blob/243051eb55c4022ff9b6bc6fd5d64462fab3efef/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L169) this extension point triggers for 
[rosidl_typesupport_c](https://github.com/ros2/rosidl_typesupport/blob/45f5a0ad4d9eaa5fc9303be2633a1e306b84026f/rosidl_typesupport_c/cmake/rosidl_typesupport_c_generate_interfaces.cmake#L70) and rosidl_typesupport_cpp](https://github.com/ros2/rosidl_typesupport/blob/45f5a0ad4d9eaa5fc9303be2633a1e306b84026f/rosidl_typesupport_cpp/cmake/rosidl_typesupport_cpp_generate_interfaces.cmake#L70) each of which call `get_used_typesupports` for their respective typesupports type.

--

So without the resources provided by rmw_connext_cpp and rmw_opensplice_cpp the only typesupports being included are the introspection typesupports which follow the codepath [here](https://github.com/ros2/rosidl_typesupport/blob/45f5a0ad4d9eaa5fc9303be2633a1e306b84026f/rosidl_typesupport_c/cmake/get_used_typesupports.cmake#L65-L67) which is evident in the build log from lines like

```
20:44:51 -- Using single matching rosidl_typesupport_c: rosidl_typesupport_introspection_c
... other output elided ...
20:44:51 -- Using single matching rosidl_typesupport_cpp: rosidl_typesupport_introspection_cpp
```
(This log was excerpted from [Bbin_UB64__builtin_interfaces 10](http://build.ros2.org/view/Bbin_uB64/job/Bbin_uB64__builtin_interfaces__ubuntu_bionic_amd64__binary/10/))

--

We need to include rmw_fastrtps_cpp in the injected dependencies because all build dependencies are installed at once and if apt can satisfy the rmw_implementation dependency with connext or opensplice it won't attempt to install fastrtps.

--

rmw_implementation itself can probably be dropped from the injected list as it will be pulled in by rosidl_default_generators via rosidl_generator_py.

dirk-thomas:
Thanks for the detailed explanation. I can see where our dependency is bypassed by the requirement on the ament index providing information about available rmw impl. Maybe a cleaner approach would be to not query the rmw impl. and derive their typesupports from them but instead just query the available typesupport directly (no matter if a rmw impl. using that typesupport is present). That would maintain the intended dependency hierarchy.

mikaelarguedas:
After talking offline with @dirk-thomas, let's merge this for now and ticket the following work of decoupling typesupport, rmw_implementations and ament_index to be addressed pose bouncy.

Note that all message packages will need to be rereleased with this bloom change

PR #485, opened 2018-08-29
Travis CI: Add Python 3.7 to the testing

cclauss:
Add Python 3.7 to the testing in alignment with travis-ci/travis-ci#9069

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ python:
   - "2.7"
   - "3.4"
   - "3.6"
+matrix:
+  include:
+    - python: "3.7"
+      dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
+      sudo: required  # required for Python 3.7 (travis-ci/travis-ci#9069)
 # command to install dependencies
 install:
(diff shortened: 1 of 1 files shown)


cclauss:
__async__ and __os.errno__ are two examples that I have been running into but https://docs.python.org/3/whatsnew/3.7.html has others.  I am OK if you do not want to add complexity to the testing.

wjwwood:
@cclauss well, since we're not using `asyncio` then I'd say we don't really need it. I wasn't aware of `os.errno` changing, but I don't think we're using that either.

I think I would decline this until either the 3.7 workaround is no longer required (`sudo: true` make it slower) or we're actively using some feature that's changed in 3.7. Also, none of our ubuntu or debian platforms use 3.7 yet, though I recognize others are using it already (I'm using it on my mac).

cclauss:
Understood.  FYI, I have not seen the slowness of sudo on Travis.  The 3.7 job usually start first and complete in less time than the Docker jobs.

wjwwood:
That may have changed at some point, but for a long time they were discouraging the use of the `sudo: true` option. I still would prefer to reevaluate this when the 3.7 issues have been resolved upstream (if that ever happens).

PR #486, opened 2018-09-15
Another bitesize PR for gitlab compatibility

DLu:
Two things: 
 * Change so that the file at `oauth_config_path` i.e. `~/.config/bloom` is overwritten as it is being written to. Appending to the file results in invalid json. 
 * Save the name of the sever as `server` in order to determine which type of PR to make. 

The next PR will have the actual gitlab interface.


bloom/commands/release.py
```diff
@@ -792,11 +792,16 @@ def open_pull_request(track, repository, distro, interactive, override_release_r
...
     version = updated_distribution_file.repositories[repository].release_repository.version
     updated_distro_file_yaml = yaml_from_distribution_file(updated_distribution_file)
-    # Determine if the distro file is hosted on github...
-    base_org, base_repo, base_branch, base_path = get_gh_info(get_distribution_file_url(distro))
-    if None in [base_org, base_repo, base_branch, base_path]:
+
+    # Determine where the distro file is hosted...
+    distro_url = get_distribution_file_url(distro)
+    base_org, base_repo, base_branch, base_path = get_gh_info(distro_url)
+    if None not in [base_org, base_repo, base_branch, base_path]:
```
nuclearsandwich: I know you've got later changes coming that might influence this but I preferred the previous expression of this condition:

```
if None in [base_org, base_repo, base_branch, base_path]:
    warning("Automated pull request only available via github.com")
    return
server = 'http://github.com'
```

makes me worry less that `server` could be undefined when used later.

DLu: How would you feel about a more drastic change that looked like this: https://github.com/ros-infrastructure/bloom/commit/44e0fa5185b617c12e69a438a5e4384fa9652c23

(not a complete change, but gives the idea)

nuclearsandwich: > How would you feel about a more drastic change that looked like this: [44e0fa5](https://github.com/ros-infrastructure/bloom/commit/44e0fa5185b617c12e69a438a5e4384fa9652c23)

I think moving in that direction makes sense. Short of an UpstreamPullRequest factory which spits out provider-specific objects with a common interface I think being able to delegate using a conditional that checks server domain is reasonable. I could also see wanting to extend the bloom config format so that mycode.lol can be configured to be a GitLab instance or BitBucket Server or whatever.

DLu:
@nuclearsandwich My sincere apologies for causing the extra work for you. I had worked on the original PR more carefully, and did not give adequate attention to the PR Feedback. It was an improper balance of trying to respond to feedback quickly and doing the due dilligence. 

With this latest set of commits, I have merged your change and the upstream changes, and fixed one minor bug. I have tested it with both a github and gitlab repo, and it works in the first case and properly complains in the second. 

Again, I'm sorry, and will try to be more thorough in future PRs.

nuclearsandwich:
Thanks for the iterations @DLu. Looking forward to releasing the GitLab compatibility in the future.

PR #481, opened 2018-07-14
WIP: File Pull Requests with Python 3

allenh1:
# NOTE: THIS IS TERRIBLE CODE
## (for now)

Very much a work in progress, with some rather interesting results (see ros/rosdistro #18480).

Most of the issue, from my observation, is just string related. This code is terrible, since it just copies and changes them (and probably also breaks python2).

PR #489, opened 2018-09-29
Fix python 3 support for pull request generation.

nuclearsandwich:
Supercedes #481 to fix GitHub pull request generation on Python3.

@allenh1 can you give this a whirl when you have a chance and see if it works for you. I've got it generating pull requests in a test environment from both python3 and python2.

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -423,10 +423,24 @@ def get_relative_distribution_file_path(distro):
 
 
 def generate_release_tag(distro):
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
I've rebased this branch onto 0.6.9. Hopefully it still works. The patch was clean...

allenh1:
Woops -- this fell into the back of my mind. I'll put this patch on the gentoo overlay and test it in a bit.

allenh1:
@nuclearsandwich looks good to me! Patch is up for Gentoo on the ros-overlay.

PR #490, opened 2018-09-29
Gitlab library

DLu:
This PR has not been tested in isolation, but is seemingly valid Python. 

This is mostly to debate whether or not I'm allowed to use Python requests. The conversation was previously [here](https://github.com/ros-infrastructure/bloom/pull/458#issuecomment-402455261) where @nuclearsandwich said they were warming to the new dependency, but I wanted to reopen that conversation.

### DIFF
--- a/bloom/gitlab.py
+++ b/bloom/gitlab.py
@@ -1,0 +1,90 @@
+import requests
+
+
+class GitlabException(Exception):
+    def __init__(self, msg, code=None):
+        if code:
+            msg = "{msg}: {code}".format(**locals())
+        super(GitlabException, self).__init__(msg)
+
+
+class GitlabAuthException(GitlabException):
+    def __init__(self, msg, code=None):
+        super(GitlabAuthException, self).__init__(msg, code)
+
+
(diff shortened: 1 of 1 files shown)


DLu:
I still need to do the requested Trusty testing on this, but please check that I added the dependencies properly.

DLu:
@nuclearsandwich I finally had some time to install trusty and check things here. requests >= 2.2 is valid there, and the CI is now passing.

nuclearsandwich:
@DLu I know I mentioned privately that I had no reservations about this PR. And that's still largely the case, but what was the reason for removing the requests module from setup.py? I believe it will still be needed there for when bloom is installed via pip / from source.

DLu:
Sorry for the delay. 

I'm not sure how setup.py really works, but with requests module in setup.py, [it broken the CI](https://travis-ci.org/ros-infrastructure/bloom/jobs/440304855#L779)

Without it, it passes. 

:man_shrugging:

DLu:
@nuclearsandwich Is this something there is still interest in merging in?

PR #493, opened 2018-11-07
Update minimum version of rosdistro to support rosdistro index v4.

nuclearsandwich:
For now we don't make any use of the new distribution data in index v4.
This is a minimal update to fix failures with bloom due to the version
restrictions.

This PR has been tested locally.

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -120,8 +120,8 @@ from vcstools.vcs_abstraction import get_vcs_client
 
 try:
     import rosdistro
-    if parse_version(rosdistro.__version__) < parse_version('0.4.0'):
-        error("rosdistro version 0.4.0 or greater is required, found '{0}' from '{1}'."
+    if parse_version(rosdistro.__version__) < parse_version('0.7.0'):
+        error("rosdistro version 0.7.0 or greater is required, found '{0}' from '{1}'."
               .format(rosdistro.__version__, os.path.dirname(rosdistro.__file__)),
               exit=True)
 except ImportError:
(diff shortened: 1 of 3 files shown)


dirk-thomas:
What is the reason to require the new `rosdistro` version?

nuclearsandwich:
> What is the reason to require the new `rosdistro` version?

If we're going to support the index v4 we should require a rosdistro index version that also supports it.

dirk-thomas:
> If we're going to support the index v4 we should require a rosdistro index version that also supports it.

I don't understand the rational here. In the original description you explicitly stated: 

> For now we don't make any use of the new distribution data in index v4.
> This is a minimal update to fix failures with bloom due to the version restrictions.

nuclearsandwich:
> I don't understand the rational here. In the original description you explicitly stated:

We're not making explicit use of the v4 data but that doesn't mean bloom won't attempt to parse the v4 file and raise an error if the version of rosdep is not able to parse it.

This was an ASAP release per your request and I think bumping the required rosdistro version is a reasonable thing to do to avoid future "gotchas" where bloom is updated but rosdistro is not.

If you would like to test an alternative change against rosdep 0.4.0 and verify it you're welcome to do so but please be cognizant of the QA work that gets put into bloom releases rather than assuming that changing a 3 to a 4 and hoping that CI passes is sufficient to avoid issues popping up after release.

dirk-thomas:
> that doesn't mean bloom won't attempt to parse the v4 file and raise an error if the version of rosdep is not able to parse it.

> avoid future "gotchas" where bloom is updated but rosdistro is not.

> assuming that changing a 3 to a 4 and hoping that CI passes is sufficient to avoid issues popping up after release.

With the changes in v4 in mind I don't see any technical possibility how that should result in incompatibilities. If you do please describe in which concrete scenario that would the case.

Anyway I will this up to you to decide if you want to bump the version dependency anyway.

PR #511, opened 2019-01-26
Require rosdep 0.15.0 in order to handle conditions in dependencies.

nuclearsandwich:
As of 0.15.0 rosdep evaluates package format 3 conditions before resolving dependencies (https://github.com/ros-infrastructure/rosdep/pull/654). In order for bloom to support them as well it needs that version of rosdep.


nuclearsandwich:
Merging on my own authority because it is Saturday.

PR #519, opened 2019-03-01
Evaluate dependency conditions

nuclearsandwich:
Noodling on this approach in order to try and unblock @AAlon.

Quite aside from the clear need to factor the condition check into a function, a few other things stand out as worth considering before merging:
* For consistent results, maybe we could get our context from the bloom tracks.yaml rather than the environment
* I used comprehensions to filter conditions on lists of dependencies roughly in place but it's worth looking more closely (at a better time of day) to see if there's any reason not to check the evaluated conditions lazily after the list of keys to evaluate is generated rather than doing it on each list as its constructed.

An additional TODO is to check that the build_type condition is also checked and honored.

### DIFF
--- a/bloom/generators/common.py
+++ b/bloom/generators/common.py
@@ -40,6 +40,8 @@ from bloom.logging import debug
 from bloom.logging import error
 from bloom.logging import info
 
+from bloom.rosdistro_api import get_distribution_type
+
 from bloom.util import code
 from bloom.util import maybe_continue
 from bloom.util import print_exc
@@ -115,6 +117,21 @@ def resolve_more_for_os(rosdep_key, view, installer, os_name, os_version):
                                              default_os_installer)
     assert inst_key in os_installers
     return installer.resolve(rule), inst_key, default_os_installer
+
+
+def package_conditional_context(ros_distro):
+    distribution_type = get_distribution_type(ros_distro)
+    if distribution_type == 'ros1':
+        ros_version = '1'
+    elif distribution_type == 'ros2':
+        ros_version = '2'
+    else:
+        error("Bloom cannot cope with distribution_type '{0}'".format(
+            distribution_type), exit=True)
+    return {
+            'ROS_VERSION': ros_version,
+            'ROS_DISTRO': ros_distro,
+            }
 
 
 def resolve_rosdep_key(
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
I put this in ready because I was curious where the Travis-CI output was. It seems this draft-state PR biz interferes with Travis running.

nuclearsandwich:
@AAlon I tweaked this so it gets the conditonal context from the rosdistro index rather than the local environment as I think that's a more reasonable primary source. Details in 0cdcd09.

It behaved sensibly for my canary package but I'm sure it could do with more rigorous field testing.

AAlon:
Thanks a lot for the quick fix! we'll retry the release in the next couple of days and will let you know how it goes.

nuclearsandwich:
The CI failure is due to my rpm implementation apparently being incomplete. I'll be able to address it before the fix is released.

AAlon:
> The CI failure is due to my rpm implementation apparently being incomplete. I'll be able to address it before the fix is released.

The RPM issue was blocking from actually going forward with the release as the tracks update stage is performed after RPM generation. The changes don't look specific to debian/rpm so I went ahead and copied the same changes you've made in `debian/generator.py` to `rpm/generator.py` (might be better to extract it into `common.py`...). Seems like it's working :-)
kinetic: https://github.com/ros/rosdistro/pull/20614
melodic: https://github.com/ros/rosdistro/pull/20615

JWhitleyAStuff:
Tested on hybrid package `ml_classifiers`. Seems to work correctly with changes suggested by @AAlon:

Kinetic: https://github.com/ros/rosdistro/pull/20839
Melodic: https://github.com/ros/rosdistro/pull/20840
Crystal: https://github.com/ros/rosdistro/pull/20841

PR for changes suggested by @AAlon: #527

nuclearsandwich:
@cottsay if you have time to review the changes tomorrow (hopefully today before you read this) then that's great. If not, no worries as I'm pretty certain that this won't break existing distros supporting Fedora. It should bring full conditional dependency support to Fedora but testing the ROS 2 side fails since the ament build types aren't yet supported.

nuclearsandwich:
> `ROS_PYTHON_VERSION`

I haven't added this yet because I'm not sure where we ought to get it from. We set the ROS_VERSION for bloomed packages from the distribution type but we have no such info for the python version. I could see it being added either to the index or to the distribution.yaml files as part of the platform metadata but I've been hesitant to pull from the local environment.

Issue #518, opened 2019-02-27
debian generator documentation is incomplete


VictorLamoine:
Help command (https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/generate_cmd.py#L64)
```bash
$ bloom-generate debian --help
usage: bloom-generate debian [-h]
                             [--place-template-files | --process-template-files]
                             [--os-name OS_NAME] [--os-version OS_VERSION]
                             [--ros-distro ROS_DISTRO] [--native] [-d]
                             [--version] [--unsafe]
                             [package_path]

Generates debian packaging files for a catkin package

positional arguments:
  package_path          path to or containing the package.xml of a package

optional arguments:
  -h, --help            show this help message and exit
  --place-template-files
                        places debian/* template files only
  --process-template-files
                        processes templates in debian/* only
  --os-name OS_NAME     OS name, e.g. ubuntu, debian
  --os-version OS_VERSION
                        OS version or codename, e.g. precise, wheezy
  --ros-distro ROS_DISTRO
                        ROS distro, e.g. indigo, kinetic, lunar, melodic (used
                        for rosdep)
  --native              generate native package

global:
  -d, --debug           enable debug messages
  --version             prints the bloom version
  --unsafe              Makes bloom faster, but if there is an error then you
                        could run into trouble.
```

However this generator seems to be able to handle more arguments, for example `--install-prefix`.
https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/generator.py#L574-L592

Am I misunderstanding how the generator works or is the `--help` documentation out-of-date?

nuclearsandwich:
The changes contributed in #526 help here but the `bloom-generate debian` subcommand doesn't have an install-prefix option, though the `bloom-generate rosdebian` subcommand [does]

As mentioned in https://github.com/ros-infrastructure/bloom/pull/526#issuecomment-481870938, the `bloom-generate ...` and `git bloom-generate ...` commands are distinct and have different sets of arguments.

To use the `--install-prefix` option you would need to use the `git bloom-generate ...` command instead.

Alternatively if you wanted to open a pull request adding support for the option, I would review it. Although I haven't checked to see if there's a reason it's not currently implemented for these commands.

I am going to close this issue, having addressed your question but please re-open it if this answer seems incomplete to you.

PR #526, opened 2019-04-01
fix issue 518

lennonwoo:
The reason why help doc is uncompleted is that wrong prepare_argument function given in [this line](https://github.com/ros-infrastructure/bloom/blob/53f129595d254e4406432c9ef63bf451d252d751/bloom/generators/rosdebian.py#L165). So in this PR, I changed it to point to the right function.

### DIFF
--- a/bloom/commands/generate.py
+++ b/bloom/commands/generate.py
@@ -67,7 +67,7 @@ def create_subparsers(parser, generator_cmds):
     for generator_cmd in generator_cmds:
         desc = load_generator_description(generator_cmd)
         cmd_parser = subparser.add_parser(desc['title'], description=desc['description'])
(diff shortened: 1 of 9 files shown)


nuclearsandwich:
Thanks for helping out. I'm very grateful that you saw an open issue and opened a pull request.

I do not think the wrong prepare_arguments function is used. The bloom-generate command and the git-bloom-generate commands are different and have different sets of arguments. The `prepare_arguments` in the current description is explicitly imported from the "parent" generate_cmd https://github.com/ros-infrastructure/bloom/blob/3f7df8faff7b34cb62f6048204e7be90e4e13d0b/bloom/generators/rosdebian.py#L43

I will update #518 clarifying as well. If you would like to open a separate PR that adds support for an --install-prefix argument to the bloom-generate command I would be happy to review it.

PR #527, opened 2019-04-09
Applying changes to rpm/generator.py suggested by @AAlon.

JWhitleyAStuff:
Tested as working on hybrid package [`ml_classifiers`](https://github.com/astuff/ml_classifiers).


nuclearsandwich:
@cottsay this looks cromulent to me but if you have some test packages you can use to put it through its paces I'd be appreciative.

nuclearsandwich:
I've tested this with the same conditional [canary package](https://github.com/nuclearsandwich/canary-condition) I tested #519 with initially. Merging into #519 to allow final review there.

PR #537, opened 2019-05-14
Improve error handling for package conditions and index files.

### LINKED ISSUE #536: bloom-release 0.8.0 always fails with "Bloom cannot cope with distribution_type 'None'"
We have an internal GitLab server and a private ROS build farm that we release packages to using bloom.  After upgrading to bloom 0.8.0, attempting to release any repository fails like so (actual repository and package names have been removed):

```
$ bloom-release -r kinetic example_repo
==> Fetching 'example_repo' repository from 'git@gitlab-server:example/example_repo-release.git'
Cloning into '/tmp/tmpVYnP9e'...
remote: Enumerating objects: 6871, done.
remote: Total 6871 (delta 0), reused 0 (delta 0)
Receiving objects: 100% (6871/6871), 2.16 MiB | 0 bytes/s, done.
Resolving deltas: 100% (2651/2651), done.
Checking connectivity... done.
==> Testing for push permission on release repository
==> git remote -v
origin  git@gitlab-server:example/example_repo-release.git (fetch)
origin  git@gitlab-server:example/example_repo-release.git (push)
==> git push --dry-run
Everything up-to-date
==> Releasing 'iop_bridge' using release track 'kinetic'
==> git-bloom-release kinetic
Processing release track settings for 'kinetic'
Checking upstream devel branch for package.xml(s)
Cloning into '/tmp/tmpZBm7_v/upstream'...
remote: Enumerating objects: 3723, done.
remote: Counting objects: 100% (3723/3723), done.
remote: Compressing objects: 100% (990/990), done.
remote: Total 3723 (delta 2612), reused 3601 (delta 2526)
Receiving objects: 100% (3723/3723), 1.64 MiB | 0 bytes/s, done.
Resolving deltas: 100% (2612/2612), done.
Checking connectivity... done.
Looking for packages in 'master' branch... found 7 packages.
Detected version '7.2.1' from package(s): [... packages ...]

Executing release track 'kinetic'
==> bloom-export-upstream /tmp/tmpZBm7_v/upstream git --tag 7.2.1 --display-uri git@gitlab-server:example/example_repo.git --name upstream --output-dir /tmp/tmpM7UgTp
Checking out repository at 'git@gitlab-server:example/example_repo.git' to reference '7.2.1'.
Exporting to archive: '/tmp/tmpM7UgTp/upstream-7.2.1.tar.gz'
Cloning into '/tmp/tmphzkP1P'...
warning: --depth is ignored in local clones; use file:// instead.
done.
md5: 803dc81cec47674755cbc4f2c72c2574

==> git-bloom-import-upstream /tmp/tmpM7UgTp/upstream-7.2.1.tar.gz  --release-version 7.2.1 --replace
The latest upstream tag in the release repository is 'upstream/7.2.0'.
Importing archive into upstream branch...
Creating tag: 'upstream/7.2.1'
I'm happy.  You should be too.

==> git-bloom-generate -y rosrelease kinetic --source upstream -i 1
Releasing packages: [... packages ...]
...

==> git-bloom-generate -y rosdebian --prefix release/kinetic kinetic -i 1 --os-name ubuntu
Generating source debs for the packages: [... packages ...]
Debian Incremental Version: 1
Debian Distributions: ['xenial']
Releasing for rosdistro: kinetic

Pre-verifying Debian dependency keys...
Running 'rosdep update'...
Bloom cannot cope with distribution_type 'None'
<== Error running command '['/usr/bin/git-bloom-generate', '-y', 'rosdebian', '--prefix', 'release/kinetic', 'kinetic', '-i', '1', '--os-name', 'ubuntu']'
Release failed, exiting.
```

If I downgrade the `python-bloom` package to version `0.7.2-1`, everything works again.

Do you know why this might be broken?

nuclearsandwich:
Closes #536 with some improved error messaging and behavior.

1. In order to release packages with format 3 package.xml files, a v4 rosdistro index is required to provide sufficient context for evaluating package conditions.
To make it clearer that the problem is the index version and not an invalid / unknown distribution_type, an explicit version check and dedicated error message has been added.

2. For package formats 1 and 2 conditions are evaluated with an empty context so that downstream code will correctly see all fields without conditions as True.

### DIFF
--- a/bloom/generators/common.py
+++ b/bloom/generators/common.py
@@ -120,6 +120,13 @@ def resolve_more_for_os(rosdep_key, view, installer, os_name, os_version):
 
 
 def package_conditional_context(ros_distro):
+    """
+    Creates a dict containing the conditional evaluation context for
+    package.xml format three packages.
+
+    :param ros_distro: The codename of the rosdistro to generate context for.
+    :returns: dict defining ROS_VERSION and ROS_DISTRO.
+    """
     distribution_type = get_distribution_type(ros_distro)
     if distribution_type == 'ros1':
         ros_version = '1'
--- a/bloom/generators/common.py
+++ b/bloom/generators/common.py
@@ -144,6 +144,7 @@ def package_conditional_context(ros_distro):
             'ROS_DISTRO': ros_distro,
             }
 
+
 def evaluate_package_conditions(package, ros_distro):
     """
     Evaluates a package's conditional fields.
--- a/bloom/generators/common.py
+++ b/bloom/generators/common.py
@@ -41,6 +41,7 @@ from bloom.logging import error
 from bloom.logging import info
 
 from bloom.rosdistro_api import get_distribution_type
+from bloom.rosdistro_api import get_index
 
 from bloom.util import code
 from bloom.util import maybe_continue
@@ -127,6 +128,9 @@ def package_conditional_context(ros_distro):
     :param ros_distro: The codename of the rosdistro to generate context for.
     :returns: dict defining ROS_VERSION and ROS_DISTRO.
     """
+    if get_index().version < 4:
+        error("Bloom requires a version 4 or greater rosdistro index to support package format 3.", exit=True)
+
     distribution_type = get_distribution_type(ros_distro)
     if distribution_type == 'ros1':
         ros_version = '1'
--- a/bloom/generators/common.py
+++ b/bloom/generators/common.py
@@ -147,7 +147,7 @@ def package_conditional_context(ros_distro):
 
 def evaluate_package_conditions(package, ros_distro):
     """
-    Evaluates a package's conditional fields.
+    Evaluates a package's conditional fields if it supports them.
 
     :param package: The package to be evaluated.
     :param ros_distro: The codename of the rosdistro use for context.
@@ -156,9 +156,7 @@ def evaluate_package_conditions(package, ros_distro):
     # Conditional fields were introduced in package format 3.
     # Earlier formats should have their conditions evaluated with no context so
     # the evaluated_condition is set to True in all cases.
-    if package.package_format < 3:
-        package.evaluate_conditions({})
-    else:
+    if package.package_format >= 3:
         package.evaluate_conditions(package_conditional_context(ros_distro))
 
 
--- a/bloom/generators/common.py
+++ b/bloom/generators/common.py
@@ -144,6 +144,22 @@ def package_conditional_context(ros_distro):
             'ROS_DISTRO': ros_distro,
             }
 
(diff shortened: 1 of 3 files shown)


bloom/generators/common.py
```diff
@@ -133,6 +144,22 @@ def package_conditional_context(ros_distro):
...
+
+    :param package: The package to be evaluated.
+    :param ros_distro: The codename of the rosdistro use for context.
+    :returns: None. The given package will be modified.
+    """
+    # Conditional fields were introduced in package format 3.
+    # Earlier formats should have their conditions evaluated with no context so
+    # the evaluated_condition is set to True in all cases.
+    if package.package_format < 3:
+        package.evaluate_conditions({})
```
dirk-thomas: That doesn't seem to make sense. Why would this be needed if the format doesn't even have conditions?

nuclearsandwich: I had initially done this so that evaluated_condition was true on all cases so that the generators would not need to duplicate the package version check.

I've since looked at [a condition check in catkin_pkg](https://github.com/ros-infrastructure/catkin_pkg/blob/7b889b2153485943ad3eb19e824059c5d04edb2d/src/catkin_pkg/package.py#L155) and switched Bloom to using the same `evaluated_condition is not False` check to allow for an evaluated_condition of None.

I think there's some risk to doing this in general because until evaluate_conditions() is invoked all evaluated_condition values are `None`. So attention must be paid to ensure that evaluate_conditions is always invoked when needed.

bloom/generators/common.py
```diff
@@ -134,6 +145,21 @@ def package_conditional_context(ros_distro):
...
+
+    :param package: The package to be evaluated.
+    :param ros_distro: The codename of the rosdistro use for context.
+    :returns: None. The given package will be modified.
+    """
+    # Conditional fields were introduced in package format 3.
+    # Earlier formats should have their conditions evaluated with no context so
+    # the evaluated_condition is set to True in all cases.
+    if package.package_format >= 3:
+        package.evaluate_conditions(package_conditional_context(ros_distro))
```
lennonwoo: Accorint to the latest implement of [catkin_pkg](https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/package.py#L183), the package.evaluate_conditions(package_conditional_context(ros_distro)) would work when package_format equal 2. As we can see from [there](https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/package.py#L344), the package.evaluate_conditions change the method evaluated_condition of every dependence to dependence object's attribute. Consider we use this attribute lately, I think this function should be also executed when package_format equals 2.

Issue #259, opened 2014-03-13
Accelerate Bloom by using `git clone --depth 1`


thomas-moulard:
Apparently, the last Git versions allow to commit and push on top of a shallow clone.
Given that the rosdistro repository is getting larger and larger, it could be a nice improvement.

At my location, cloning rosdistro (and my own package) is by far the longest operation.

Cf (for instance) http://stackoverflow.com/questions/6941889/is-git-clone-depth-1-shallow-clone-more-useful-than-it-makes-out

[note that this version of git is quite new so Bloom would need to make sure that the installed Git is recent enough, >=1.9-rc2]

PR #538, opened 2019-06-07
use shallow clone for rosdistro repository

rhaschke:
This implements a 5-year old feature request, i.e. #259, to checkout the `rosdistro` repo in a shallow fashion. As the `rosdistro` repo incrementally grows, this feature become increasingly more important.

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -740,7 +740,7 @@ Increasing version of package(s) in repository `{repository}` to `{version}`:
                     warning("Skipping the pull request...")
                     return
                 _my_run('git checkout -b {new_branch}'.format(**locals()))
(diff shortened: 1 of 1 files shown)


mikepurvis:
Looks simple enough. Main thing is just confirming that this does the right thing when multiple interleaved PRs hit at the same time, though I suppose from Github's point of view, it should be transparent whether a PR branch came from a shallow clone or now.

rhaschke:
There shouldn't be any interactions between multiple PRs: A PR is always made w.r.t. the latest commit. That's the only thing that matters :wink:

rhaschke:
Unfortunately, I frequently get errors like
[remote rejected] ... (shallow update not allowed)
when trying to push back local changes. Probably, git cannot connect the locally checked out branch to the further progressed branch of the main repo.
Closing for now.

PR #547, opened 2019-09-04
fix #546: make debhelper --parallel

### LINKED ISSUE #546: generator templates for debian using debhelper do not support parallel builds
There is no support for parallel builds in any of the four generators that target the debian platform. More exactly, the `--parallel` option to the debhelper command within the default rule is conspicuously absent for each of the following:

* `bloom/generators/debian/templates/ament_cmake/rules.em`
* `bloom/generators/debian/templates/ament_python/rules.em`
* `bloom/generators/debian/templates/catkin/rules.em`
* `bloom/generators/debian/templates/cmake/rules.em`

realtime-neil:


realtime-neil:
ping?

tfoote:
I'm assuming that this causes debhelper to run parallel make inside. Please link to documentation of the flag. And include when this option became available, who else is using it and where. And if it's now recommended in the debian packaging guidelines. 

This is a potentially big and change. As such this is likely something that we would want to phase in based on new releases to not break existing packages that might not be safe to build in parallel and rely on the old behavior.

The first link suggests that this option is no longer required and that the environmet variable is what's necessary: https://askubuntu.com/questions/337093/how-to-run-parallel-make-with-debuild

realtime-neil:
Granted, @tfoote, for `debhelper` `compat` 10 and above, `--parallel` is the default.

> dh sequencer based packages now defaults to building in parallel (i.e. “–parallel” is default in compat 10)

-- https://nthykier.wordpress.com/2016/09/11/debhelper-10-is-now-available/

And it's good that the four `compat.em` templates all take their values from the installed debhelper version:
* https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/ament_cmake/compat.em
* https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/ament_python/compat.em
* https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/catkin/compat.em
* https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/cmake/compat.em

And yet:
```
$ for ros_distro in kinetic melodic crystal dashing; do echo "ros_distro: dashing" && docker container run --rm -it "ros:dashing" sh -c 'apt-get update >/dev/null 2>&1 && apt-cache policy debhelper'; done
ros_distro: kinetic
debhelper:
  Installed: (none)
  Candidate: 9.20160115ubuntu3
  Version table:
     10.2.2ubuntu1~ubuntu16.04.1 100
        100 http://archive.ubuntu.com/ubuntu xenial-backports/main amd64 Packages
     9.20160115ubuntu3 500
        500 http://archive.ubuntu.com/ubuntu xenial/main amd64 Packages
ros_distro: melodic
debhelper:
  Installed: (none)
  Candidate: 11.1.6ubuntu2
  Version table:
     12.1.1ubuntu1~ubuntu18.04.1 100
        100 http://archive.ubuntu.com/ubuntu bionic-backports/main amd64 Packages
     11.1.6ubuntu2 500
        500 http://archive.ubuntu.com/ubuntu bionic-updates/main amd64 Packages
     11.1.6ubuntu1 500
        500 http://archive.ubuntu.com/ubuntu bionic/main amd64 Packages
ros_distro: crystal
debhelper:
  Installed: (none)
  Candidate: 11.1.6ubuntu2
  Version table:
     12.1.1ubuntu1~ubuntu18.04.1 100
        100 http://archive.ubuntu.com/ubuntu bionic-backports/main amd64 Packages
     11.1.6ubuntu2 500
        500 http://archive.ubuntu.com/ubuntu bionic-updates/main amd64 Packages
     11.1.6ubuntu1 500
        500 http://archive.ubuntu.com/ubuntu bionic/main amd64 Packages
ros_distro: dashing
debhelper:
  Installed: (none)
  Candidate: 11.1.6ubuntu2
  Version table:
     12.1.1ubuntu1~ubuntu18.04.1 100
        100 http://archive.ubuntu.com/ubuntu bionic-backports/main amd64 Packages
     11.1.6ubuntu2 500
        500 http://archive.ubuntu.com/ubuntu bionic-updates/main amd64 Packages
     11.1.6ubuntu1 500
        500 http://archive.ubuntu.com/ubuntu bionic/main amd64 Packages
```

Even with the assumption that `debhelper` version 10+ automagically parallelizes the build --- an assumption that I will test in forthcoming comments --- this begs the question: What about Kinetic?

tfoote:
As I mentioned above, for something like this in general we wouldn't want to change an underlying behavior like this in a stable released distribution. As such I think that sticking with the defaults we're already using makes sense. We don't want to disrupt the already released distributions. If we were to roll this out we'd want to explicitly only enable parallelization for upcoming distributions. 

Depending on how this is ends up executing it might actually cause issues on the buildfarm. Right now the buildfarm runs multiple builds in parallel based on the assumption that each build is single threaded and won't use the whole machine. We're also significantly IO limited on the buildfarm which this will not change so I don't see it as a significant cost reduction either.

PR #552, opened 2019-10-18
Prepare 0.9.0

nuclearsandwich:
This commit should be fast-forwarded rather than merged via GitHub.

The 0.9.0 release has two significant changes: ament (ROS 2) support for RPM spec generation and support for the `ROS_PYTHON_VERSION` dependency conditional.

With the latter change, bloom will generate an error when releasing any package into a rosdistro with no `python_version` defined in the rosdistro index. This is intentional, see #551 for the rationale.


nuclearsandwich:
Holding on to add #553

nuclearsandwich:
Re-written to correct some whitespace changelog and add #553

PR #559, opened 2019-12-30
Also run tests with Python 3.7 and 3.8

sloretz:
Debian Buster uses Python 3.7 and Ubuntu Focal will use Python 3.8


sloretz:
Python 3.8 CI failure looks like it would be solved by https://github.com/ros-infrastructure/rospkg/pull/189

PR #561, opened 2020-01-14
Depend on rosdistro 0.8.0 to solve problems with test_abi key

j-rivero:
[Some](https://answers.ros.org/question/341427) [users](#issuecomment-573302415) reported problems with bloom after the introduction of the `test_abi` key in https://github.com/ros-infrastructure/rosdistro/pull/147. In order to be sure that upgrades of bloom use the correct rosdistro, this PR changes dependencies to pull the `0.8.0` version which implements the `test_abi` parsing.

After merging we probably need to create a new release of bloom.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ install_requires = [
     'python-dateutil',
     'PyYAML',
     'rosdep >= 0.15.0',
-    'rosdistro >= 0.7.5',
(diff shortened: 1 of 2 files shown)


j-rivero:
I thin that the problem in CI has nothing to do with this PR.

tfoote:
CI fix proposed in #567

PR #567, opened 2020-01-16
Resolve PyYAML Python 3.4 support in CI

tfoote:
CI is failing on python 3.4 because the new version of PyYAML dropped support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ python:
   - "3.6"
 # command to install dependencies
 install:
-  - pip install PyYAML==5.2 # Forcing PyYAML 5.2 while we retain Python 3.4 support PyYAML 5.3 and higher does not support python 3.4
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,11 +1,12 @@
 language: python
 python:
   - "2.7"
-  - "3.4"
+  - "3.4"  # When support for 3.4 is removed unpin the PyYAML version below.
   - "3.6"
 # command to install dependencies
 install:
-  - pip install PyYAML argparse rosdep vcstools catkin-pkg python-dateutil setuptools
+  - pip install PyYAML==5.2 # Forcing PyYAML 5.2 while we retain Python 3.4 support PyYAML 5.3 and higher does not support python 3.4
```
dirk-thomas: The `PyYAML` version shouldn't be pinned for Python version != 3.4. See dirk-thomas/vcstool#117 for an example.

PR #571, opened 2020-01-27
add debinc to allow our version to override the upstream version

wjwwood:
What will we need to do after merging this? Rerun our `deb` and `deb3` jobs? Or do we need a new release entirely?


dirk-thomas:
It needs a new patch release - otherwise the regenerated tarball will conflict. We had that in the past.

PR #579, opened 2020-03-27
Print devel branch for version detect

bmagyar:
I had an issue this morning where I couldn't figure out why bloom was picking the previous version for re-releasing, instead of the new one. Found that for some reason the branch to pick melodic releases was still set to kinetic. One can of course go and investigate this in the release repos but in my opinion it's much quicker and cleaner if `bloom` tells us which branch it's determining the new version from.


bmagyar:
Do we need someone else to review this?

PR #581, opened 2020-04-07
Add non-interactive option to git-bloom-release command.

nuclearsandwich:
This adds a noninteractive option to git-bloom-release. Most of the time
this command is often non-interactive but several fields in the bloom
config can be set to :{ask} which would normally block with a prompt.

With this patch an error message will be raised instead if the :{ask}
option is used.


nuclearsandwich:
> Is there a way we can 'forward' the non-interactive flag to the generators so that we don't need to specify it twice?

Some of the other bloom arguments that apply to multiple commands and subcommands are exposed via environment variables. I think we could consider a BLOOM_NON_INTERACTIVE=1 variable in the future.

PR #583, opened 2020-04-30
Update changelog urls which were pointing to a fork.

nuclearsandwich:
I think this might have been an oversight when updating the changelog? The fork urls all 404 since the relevant issues/PRs are on the primary fork.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,15 +4,15 @@
 
 0.9.2 (2020-03-02 13:00:00 -0800)
 ---------------------------------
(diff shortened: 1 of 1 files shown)


cottsay (approved):
Oof. I wonder if we can somehow make noise about this in the changelog tool...

Thanks for fixing it.

nuclearsandwich:
> Oof. I wonder if we can somehow make noise about this in the changelog tool...

There's a changelog tool? I've been writing them by hand like a fool.

PR #584, opened 2020-04-30
Fix interactive prompt when running non-interactively.

nuclearsandwich:
The changes in #577 restored an interactive prompt that wasn't
functioning when #581 was being worked on. This checks for interactivity
before running the 'maybe_continue' prompt.


nuclearsandwich:
@cottsay and @tfoote I tagged both of you but I only need review from whoever gets here first. So whoever does review I humbly ask you to remove the other reviewer.

PR #598, opened 2020-06-25
Enable version specific patches

### LINKED ISSUE #597: Given patches path 'melodic/:{version}' does not exist in bloom branch.
The `patches` key in `tracks.yaml` doesn't support templating the version

sevenbill:
Adding template support for `patches` in tracks.yaml
```
    patches: melodic/:{version}
```

Fixes #597


nuclearsandwich:
Do you have an example of a release repository that is using this patch?

From the conversation in #597 it isn't clear that this actually solves any issues for you.

PR #600, opened 2020-08-04
Solve shlibdeps errors in REP136 packages that use GNUInstallDirs.

jlblancoc:
I want to discuss whether this patch is acceptable or, on the contrary, packages are enforced in ROS *not* to use multiarchs lib prefixes. In the latter case, it would be great to have this documented somewhere, if it's not already. 
Thanks!

Pros: Any 3rd party library, well-behaved in the Debian sense of implementing multiarch prefixes, would be automatically release-able via bloom w/o changes...

Cons: none?

See: 

- https://www.ros.org/reps/rep-0136.html
- https://answers.ros.org/question/356962/should-cmake_install_libdir-be-ignored-in-rep136-packages/
- https://github.com/MRPT/mrpt/issues/1076
- Maybe related? https://answers.ros.org/question/290073/problem-building-with-non-installed-dependencies/
- Maybe related? https://www.ros.org/reps/rep-0122.html#lib

### DIFF
--- a/bloom/generators/debian/templates/cmake/rules.em
+++ b/bloom/generators/debian/templates/cmake/rules.em
@@ -18,6 +18,9 @@ export PKG_CONFIG_PATH=@(InstallationPrefix)/lib/pkgconfig
 # 	https://github.com/ros-infrastructure/bloom/issues/327
 export DEB_CXXFLAGS_MAINT_APPEND=-DNDEBUG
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@cottsay @nuclearsandwich Friendly ping from the author.

nuclearsandwich (approved):

PR #607, opened 2020-09-29
update email addresses and remove 404 download url

wjwwood:
I updated the email addresses and removed the download url because from what I can tell we just shouldn't set it, and the link that was there was 404.


cottsay (approved):
Thanks!

wjwwood:
I'll wait for @tfoote to approve since I changed his email too.

tfoote (approved):

PR #625, opened 2021-03-24
Rewire the typesupport dependencies for post-Foxy.

clalancette:
Up to foxy, the connext RMW required typesupport packages.
Starting in Galactic, the new connext RMW no longer requires
those packages.  Revamp the logic here so that we only add
those connext typesupports for distros before Galactic.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

@nuclearsandwich @cottsay This is part of https://github.com/rticommunity/rmw_connextdds/issues/9 .  I'm not quite sure how to test it, so any hints there are welcome.  Also, once we merge this we'll really need a new bloom release so that we can re-release all of the interface packages into Rolling/Galactic to remove this dependency.

### DIFF
--- a/bloom/generators/rosdebian.py
+++ b/bloom/generators/rosdebian.py
@@ -108,14 +108,20 @@ class RosDebianGenerator(DebianGenerator):
                     self.rosdistro not in ('r2b2', 'r2b3', 'ardent') and \
                     'rosidl_interface_packages' in [p.name for p in package.member_of_groups]:
                 ROS2_VENDOR_TYPESUPPORT_DEPENDENCIES = [
-                    'rosidl-typesupport-connext-c',
-                    'rosidl-typesupport-connext-cpp',
                     'rosidl-typesupport-fastrtps-c',
                     'rosidl-typesupport-fastrtps-cpp',
                 ]
+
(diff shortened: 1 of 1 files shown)


cottsay (approved):
This looks good to me.

Doing a `bloom-release` of an affected repository with the `--pretend` argument should be sufficient to prove no regression.
Manually cloning the release repo and using `git-bloom-release` would allow you to inspect the results of the local repo to show that this change had the desired behavior.

clalancette:
Thanks to a little additional help from @cottsay, I was able to validate this.  For the record, here is what I did:

```
git clone https://github.com/ros-infrastructure/bloom -b clalancette/fix-bloom-deps-connext
cd bloom
python3 -m venv bloom_venv
. bloom_venv/bin/activate
python3 setup.py develop
cd ..
git clone https://github.com/ros2-gbp/rcl_interfaces-release
git-bloom-release rolling
git checkout debian/rolling/focal/rcl_interfaces
grep "Build-Depends" debian/control
```

And I verified that the `Build-Depends` for rolling no longer contains connext typesupport.  I did the same thing for foxy and dashing, and in both cases validated that they *did* contain connext typesupport, as they should.

With that, I'm going to go ahead and merge this.  Thanks for the review.

PR #627, opened 2021-04-06
Use basic authentication with new personal access tokens.

nuclearsandwich:
New personal access tokens on GitHub are not created as oauth tokens and
are not usable with token authentication. Instead they're used as the
password for basic authentication.

More info on these new tokens is in [this GitHub blog post](https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/).

This is a minimal change to ensure that bloom can continue to work with
these new tokens.

### DIFF
--- a/bloom/github.py
+++ b/bloom/github.py
@@ -80,7 +80,10 @@ def auth_header_from_basic_auth(user, password):
 
 
 def auth_header_from_oauth_token(token):
-    return "token " + token
+    auth_str = '{0}:{1}'.format('nuclearsandwich', token)
+    if sys.version_info >= (3, 0):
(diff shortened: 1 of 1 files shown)


cottsay (approved):
Is there any documentation you can reference regarding the `ghp_` prefix that this change relies on?

nuclearsandwich:
> Is there any documentation you can reference regarding the `ghp_` prefix that this change relies on?

There's a block post here that was linked to me when I was refreshing a token for one of my hosts: https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
Because it is a blog post I didn't link it in comments but it definitely ought to go in the PR body.

PR #629, opened 2021-04-06
Add a fast check for the likely fork name.

nuclearsandwich:
The ros/rosdistro repository has accumulated many forks and the current
bloom strategy of listing them all then finding the one belonging to the
current GitHub user takes time.

Most of the time a forked repository has the same name as the parent.
This PR checks that the target repository and a user repository with the
same name (if it exists) share the same "source" and are thus in the
same fork network so pull requests can be created between them.

If the fork exists with the right name this check is now almost
instantaneous on my local internet connection so this could save a lot
of seconds when working on releases to ros/rosdistro and since it
creates only two additional API requests it should not slow down the
exhaustive search by very much.

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -679,9 +679,12 @@ Increasing version of package(s) in repository `{repository}` to `{version}`:
             # Check if a fork already exists on the user's account
 
             try:
-                # There are a lot of forks of the ros/rosdistro repository so listing those forks takes a very long time.
-                # Let's try a little shortcut by checking if the repository of the same name owned by the current user
-                # if that repository exists and is in the same fork network as the target repo let's take it.
+                # There are a lot of forks of the ros/rosdistro repository so
+                # listing those forks takes a very long time.
+                # Let's try a little shortcut by checking if the repository of
+                # the same name owned by the current user, if that repository
+                # exists and is in the same fork network as the target repo
+                # let's take it.
                 # If it is not, we still fall back to listing forks.
                 target_repo = gh.get_repo(base_info['org'], base_info['repo'])
                 target_repo_source = target_repo['full_name']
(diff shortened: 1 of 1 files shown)


cottsay (approved):
Looks like a linter check failed.

I haven't tested this, but I reviewed the code, and it makes sense to me. Seems like a safe improvement.

bloom/commands/release.py
```diff
@@ -679,10 +679,31 @@ def open_pull_request(track, repository, distro, interactive, override_release_r
...
 
             try:
-                repo_forks = gh.list_forks(base_info['org'], base_info['repo'])
-                user_forks = [r for r in repo_forks if r.get('owner', {}).get('login', '') == gh.username]
-                # github allows only 1 fork per org as far as I know. We just take the first one.
-                head_repo = user_forks[0] if user_forks else None
+                # There are a lot of forks of the ros/rosdistro repository so
+                # listing those forks takes a very long time.
+                # Let's try a little shortcut by checking if the repository of
+                # the same name owned by the current user, if that repository
```
clalancette, line 685: You may want to make this comment a little clearer on what you mean by the "current user".  That is, it's the Github current user, not the Unix current user or anything like that.

```suggestion
                # the same name owned by the current Github user, if that repository
```

nuclearsandwich:
> I haven't tested this, but I reviewed the code, and it makes sense to me. Seems like a safe improvement.

I didn't write tests for this because I don't think that we have GitHub API fixtures in place but these patches were motivated by the time I spent waiting for this check to run while testing #627  and #628 and I deleted my own fork a few times to test that path.

clalancette (commented):
Yes please!  This is one of the slowest part of bloom for me.  One minor suggestion inline on the comments.

bloom/commands/release.py
nuclearsandwich: Done in a790f09. Although I spelled `GitHub` with the preferred capitalization because I'm deeply programmed to do so.

clalancette: > Although I spelled `GitHub` with the preferred capitalization because I'm deeply programmed to do so.

That's fine, I made Scott change "RedHat" to "Red Hat" in the RHEL documentation for the same reasons :).

PR #633, opened 2021-04-16
Fix token authentication error.

### LINKED ISSUE #632: Failed to open pull request, even with github access token defined
I am attempting to bloom `https://github.com/ros-perception/vision_msgs` into `rolling`, but the final pull request into `rosdistro` is failing with the following error:
```
Failed to open pull request: GithubException - HTTP Error 404: Not Found (https://api.github.com/repos/ros/rosdistro/pulls): None
```

Full output is here: [bloom-attempt-stdout.txt](https://github.com/ros-infrastructure/bloom/files/6311270/bloom-attempt-stdout.txt)

Bloom version:
```
$ bloom-release --version
0.10.6
```

Github personal access token:
```
$ cat ~/.config/bloom
{
    "github_user": "kukanani",
    "oauth_token": "ghp_<token_contents>"
}
```

Access token scopes:
Screenshot from 2021-04-14 08-59-14

Other things to note that may be helpful:
```
$ uname -a
Linux obelisk20 5.8.0-48-generic #54~20.04.1-Ubuntu SMP Sat Mar 20 13:40:25 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
* Release repo and source repo are under different users (kukanani and ros-perception, respectively)
* Tested multiple times several minutes apart, same result
* I can visit the URL that is supposedly 404ing in Firefox (showing 29 pulls at the moment)

nuclearsandwich:
The new personal access tokens from GitHub for which support was first
added in bloom in #627 use basic authentication like a username and
password rather than the OAuth token authentication used by OAuth tokens
and previous iterations of the personal access token. When adding
support for this it seems like I missed a trick and as a result the
authentication header was being sent as
`Basic b'base64encodeduserandtoken'` rather than
`Basic base64encodeduserandtoken`.

I have to do an experimental bisect to determine why this issue is only
showing up now. Perhaps it's been an issue since the jump to python3 but
since basic authentication was only ever used as a springboard for token
auth it was overlooked when reported.

I am pretty sure that this will address #632 as well as tokens being immediately rejected when entered via the prompt added in #628.

Closes #632

### DIFF
--- a/bloom/github.py
+++ b/bloom/github.py
@@ -76,7 +76,10 @@ def auth_header_from_basic_auth(user, password):
     auth_str = '{0}:{1}'.format(user, password)
     if sys.version_info >= (3, 0):
         auth_str = auth_str.encode()
-    return "Basic {0}".format(base64.b64encode(auth_str))
+    b64_encoded = base64.b64encode(auth_str)
+    if sys.version_info >= (3, 0):
+        b64_encoded = b64_encoded.decode()
+    return "Basic {0}".format(b64_encoded)
 
 
 def auth_header_from_token(username, token):


cottsay (approved):

nuclearsandwich:
I used this to create https://github.com/nuclearsandwich/rosdistro/pull/1 via the new flow.

PR #661, opened 2022-01-26
Add build dependency dh-python for --with python3

jspricke:


nuclearsandwich:
@jspricke I don't necessarily disagree with the change but I could use additional info on the motivation. Is there a link to a maintainer guide or best practice which  I'm trying to think back when the ament_python templates were first created as to why I didn't include the dh_python dependency but nothing's coming up. I wonder if it was stuffed into the ros_buildfarm environment instead of being a dependency.

I'm good with the change if including package helper tools in the build dependencies is the best practice.

jspricke:
***@***.*** I don't necessarily disagree with the change but I could use additional info on the motivation. Is there a link to a maintainer guide or best practice which  I'm trying to think back when the ament_python templates were first created as to why I didn't include the dh_python dependency but nothing's coming up. I wonder if it was stuffed into the ros_buildfarm environment instead of being a dependency.

Debian defines a list of build-essential packages, should be in the 
policy but currently I only find it here:

https://packages.debian.org/de/sid/build-essential

All other packages need to be specified in the d/control file. For 
--with python3 in d/rules you need Debhelper/Sequence/python3.pm which 
is in dh-python. The modern approach is to depend on dh-sequence-python3 
and drop the --with python3, but I guess you need to define a newer 
compat version for that.

jspricke:
This explains how it works:

https://sources.debian.org/src/build-essential/12.9/list/

And the non German link from above (sorry):
https://packages.debian.org/sid/build-essential

nuclearsandwich:
Thanks! This will be in the next release of Bloom. Packages will only adopt it as they are re-bloomed with the new version so it will be some time before this is ubiquitous.

jspricke:
Thanks! The most important thing for us is that bloom itself has the dependency so we can enable pre release testing.

PR #663, opened 2022-03-27
update distro list propmt with newer ros and ros2 distros

ijnek:

### DIFF
--- a/bloom/util.py
+++ b/bloom/util.py
@@ -263,10 +263,11 @@ _quiet = False
 _disable_git_clone = False
 _disable_git_clone_quiet = False
 _distro_list_prompt = [
-    'indigo',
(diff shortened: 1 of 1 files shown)


nuclearsandwich (commented):
@ijnek thanks for the contribution! I'm doing a quick drive-by review here but what do you think about updating this so it pulls distros from the rosdistro index whose state is currently `active`, `rolling`, or `pre-release` that way it no longer needs updating and may even work better for rosdistro forks with custom distros.

ijnek:
Sounds good, I've raised a separate PR with your suggestions.

ijnek:
Closing, superceded by #670 with suggestions by @nuclearsandwich

PR #670, opened 2022-04-20
Get non eol distros from rosdistro index

ijnek:
This is an alternative to #663 as suggested by @nuclearsandwich.


nuclearsandwich (approved):
@ijnek this is excellent! As long as this doesn't bug any tests I think this can be merged.

Thank you for the contribution!

ijnek:
@nuclearsandiwch I believe the test should pass now

ijnek:
@nuclearsandwich could you re-run the tests please?

PR #676, opened 2022-06-07
Update docs to use more recent sphinx blueprint

ijnek:
Mainly updates the docs to use a more recent sphinx blueprint, since the one being used is from 9 years ago.

Some changes: removes the versioning done in conf.py since RTD can figure this out using the git tags in the repo.

Adds sphinx-rtd-theme to requirements.txt, rather than checking if we're on RTD or not. Anyone working with the repo will need to install the requirements.txt. Maybe we could add instructions to the root README.rst about this.

Gets rid of /home/user/bloom/docs/glossary.rst: WARNING: document isn't included in any toctree, by commening out html_static_path in conf.py

Separates out general description of bloom out of index.rst into bloom.rst, so it displays in the sidebar table of contents correctly.


ijnek:
Gentle ping.

PR #684, opened 2022-08-10
Dry-run with set-upstream origin master, in case remote repo isn't initialized yet

ijnek:
Currently, a maintainer must initialize a release repository before running bloom (such as by adding a README.md). If a release repository is uninitialized, bloom will complain such as:

```sh

==> git push --dry-run
fatal: The current branch master has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin master

Cannot push to remote release repository.
```

The change in this PR makes bloom work with uninitialized remote repositories. (Hence alleviating the need to manually initialize)

Apart from this ``git push --dry-run`` command that was failing, bloom seems to be able to work with an uninitialized remote repository completely fine.

Resolves: ros2-gbp/ros2-gbp-github-org#47

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -927,7 +927,7 @@ def _perform_release(
         info(fmt("@{bf}@!==> @|@!") + str(cmd))
         subprocess.check_call(cmd, shell=True)
         # Dry run will authenticate, but not push
-        cmd = 'git push --set-upstream origin {0} --dry-run'.format(BLOOM_CONFIG_BRANCH)
+        cmd = 'git push origin {0} --dry-run'.format(BLOOM_CONFIG_BRANCH)
         info(fmt("@{bf}@!==> @|@!") + str(cmd))
         subprocess.check_call(cmd, shell=True)
     except subprocess.CalledProcessError:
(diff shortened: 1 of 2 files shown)


ijnek:
@nuclearsandwich gentle ping

ijnek:
@nuclearsandwich what's your thoughts on this?

nuclearsandwich:
Thanks for the patch!

I think this may be a bug in bloom resulting from changes in Git behavior because I have never previously had an issue with `bloom-release`ing to an empty repository.

I just pushed a change that I realize will break Python2 which we're still hanging onto until May. So I need to change that.

nuclearsandwich (approved):
Thanks for the patch! I made two changes directly.

1. Use the `BLOOM_CONFIG_BRANCH` constant rather than adding another hard-coded `'master'` if we ever change that we'll have our work cut out for us but IMO it's better not to make a bad problem worse when it's possible not to.
2. The `--set-upstream` flag isn't required to get this working, just specifying the remote target for this push is sufficient.

I'd like to give you a chance to counter-review those changes and if you have no objections we can merge this one.

ijnek (commented):
I can confirm the git push works without ``--set-upstream``, thanks for the changes!

nuclearsandwich:
oof I wish I'd updated the PR title before merging. I'm going to do a little force-push magic.. Avert your eyes if you're sensitive to that!

PR #686, opened 2022-09-09
Add Defaults for src/doc information

DLu:
If you are adding a new repository to rosdistro, it will prompt you to add Documentation and Source information, but it will not use any of the information previously used to fill in the defaults. This PR uses the `track_dict` object to populate the defaults for the fields. 

So instead of 

```
Would you like to add documentation information for this repository? [Y/n]? 
==> Looking for a doc entry for this repository in a different distribution...
No existing doc entries found for use as defaults.
Please enter your repository information for the doc generation job.
This information should point to the repository from which documentation should be generated.
VCS Type must be one of git, svn, hg, or bzr.
VCS type: git
VCS url: https://github.com/MetroRobots/ros_system_fingerprint.git
VCS version must be a branch, tag, or commit, e.g. master or 0.1.0
VCS version: ros2

```
(where I manually typed `git/https://github.com/MetroRobots/ros_system_fingerprint.git/ros2`)

we get

```
Would you like to add documentation information for this repository? [Y/n]? 
==> Looking for a doc entry for this repository in a different distribution...
No existing doc entries found for use as defaults.
Please enter your repository information for the doc generation job.
This information should point to the repository from which documentation should be generated.
VCS Type must be one of git, svn, hg, or bzr.
VCS type [git]: 
VCS url [https://github.com/MetroRobots/ros_system_fingerprint.git]: 
VCS version must be a branch, tag, or commit, e.g. master or 0.1.0
VCS version [ros2]:
```

(where `git/https://github.com/MetroRobots/ros_system_fingerprint.git/ros2` were autopopulated as the defaults and I just had to hit enter three times)

### DIFF
--- a/bloom/commands/release.py
+++ b/bloom/commands/release.py
@@ -391,7 +391,7 @@ def generate_ros_distro_diff(track, repository, distro, override_release_reposit
         defaults = defaults or {}
         while True:
             info("VCS Type must be one of git, svn, hg, or bzr.")
-            default = defaults.get('type', None)
+            default = defaults.get('type', track_dict.get('vcs_type'))
             insert = '' if default is None else ' [{0}]'.format(default)
             vcs_type = safe_input('VCS type{0}: '.format(insert))
             if not vcs_type:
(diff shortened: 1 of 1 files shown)


DLu:
What do you think @wjwwood ?

wjwwood:
I'll defer to @nuclearsandwich and @cottsay, they're really running this project these days.

I think it could be fine. I don't remember why it is like this.

nuclearsandwich (approved):
I'm into it. I don't think the interactive workflows are very well tested so I'll want to do some test releases with this off of master before this gets into a release.

Issue #678, opened 2022-06-14
PkgResourcesDeprecationWarning:  is an invalid version and will not be supported in a future release


ijnek:
Following warning spams when running ``bloom-release``:

```
/usr/lib/python3/dist-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning:  is an invalid version and will not be supported in a future release
```

This deprecation warning was [added to pypa/setuptools](https://github.com/pypa/setuptools/commit/84b1cd8e7760a9b3c984a4edfc37dc7266952266) some time last year, and it seems like all the has to be done is to use ``packaging.parse`` instead of ``pkg_resources.parse_version`` as described in [PEP 440](https://peps.python.org/pep-0440/).

I will try and get a fix in if no one gets to it first.

jbarry-bdai:
With setuptools 66.0.0 this now breaks `bloom-generate` on anything but Mac I believe:
```
Traceback (most recent call last):
  File "/usr/local/bin/bloom-generate", line 5, in <module>
    from bloom.commands.generate import main
  File "/usr/local/lib/python3.10/dist-packages/bloom/commands/__init__.py", line 42, in <module>
    from bloom.commands.update import start_updater
  File "/usr/local/lib/python3.10/dist-packages/bloom/commands/update.py", line 49, in <module>
    from bloom.logging import warning
  File "/usr/local/lib/python3.10/dist-packages/bloom/logging.py", line 63, in <module>
    _is_mac_lion_or_greater = parse_version(mac_ver()[0]) >= parse_version('10.7.0')
  File "/usr/local/lib/python3.10/dist-packages/pkg_resources/_vendor/packaging/version.py", line 266, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: ''
```
 Downgrading setuptools to 65.7.0 solved the problem.

facontidavide:
Thanks, I had this same problem. Is there any solution? I think many people might be affected!

PR #693, opened 2023-02-02
Replace pkg_resources.parse_version with packaging.version.parse

ijnek:
Resolves: #678.

Switch to packaging since use of pkg_resources is discouraged. Related: https://github.com/pypa/setuptools/issues/2497

Also:
* Uses regex to do version matching, because packaging's version parsing is a lot more strict about version numbers.
* Also, removes a check for whether a mac is Mountain Lion or greater and just changes it to a check for whether it is a mac.

### DIFF
--- a/bloom/commands/git/import_upstream.py
+++ b/bloom/commands/git/import_upstream.py
@@ -39,7 +39,7 @@ import sys
 import tarfile
 import tempfile
 
-from packaging.version import parse
+from packaging.version import parse as parse_version
 
 try:
     from urlparse import urlparse
@@ -87,7 +87,7 @@ def version_check(version):
     info(fmt("The latest upstream tag in the release repository is '@!{0}@|'."
          .format(last_tag.replace('@', '@@'))))
     # Ensure the new version is greater than the last tag
-    if parse(version) < parse(last_tag_version):
+    if parse_version(version) < parse_version(last_tag_version):
         warning("""\
 Version discrepancy:
 The upstream version '{0}' isn't newer than upstream version '{1}'.
--- a/bloom/commands/git/import_upstream.py
+++ b/bloom/commands/git/import_upstream.py
@@ -39,7 +39,7 @@ import sys
 import tarfile
 import tempfile
 
-from pkg_resources import parse_version
+from packaging import version
 
 try:
     from urlparse import urlparse
@@ -87,7 +87,7 @@ def version_check(version):
     info(fmt("The latest upstream tag in the release repository is '@!{0}@|'."
          .format(last_tag.replace('@', '@@'))))
     # Ensure the new version is greater than the last tag
-    if parse_version(version) < parse_version(last_tag_version):
+    if version.parse(version) < version.parse(last_tag_version):
         warning("""\
 Version discrepancy:
 The upstream version '{0}' isn't newer than upstream version '{1}'.
--- a/bloom/commands/git/import_upstream.py
+++ b/bloom/commands/git/import_upstream.py
@@ -39,7 +39,7 @@ import sys
 import tarfile
 import tempfile
 
-from packaging import version
+from packaging.version import parse
 
 try:
     from urlparse import urlparse
@@ -87,7 +87,7 @@ def version_check(version):
     info(fmt("The latest upstream tag in the release repository is '@!{0}@|'."
          .format(last_tag.replace('@', '@@'))))
     # Ensure the new version is greater than the last tag
-    if version.parse(version) < version.parse(last_tag_version):
+    if parse(version) < parse(last_tag_version):
         warning("""\
 Version discrepancy:
 The upstream version '{0}' isn't newer than upstream version '{1}'.
--- a/bloom/generators/release.py
+++ b/bloom/generators/release.py
@@ -53,8 +53,8 @@ from bloom.commands.git.patch.trim_cmd import trim
 
 try:
     import catkin_pkg
-    from pkg_resources import parse_version
-    if parse_version(catkin_pkg.__version__) < parse_version('0.3.8'):
+    from packaging import version
+    if version.parse(catkin_pkg.__version__) < version.parse('0.3.8'):
         warning("This version of bloom requires catkin_pkg version >= '0.3.8',"
                 " the used version of catkin_pkg is '{0}'".format(catkin_pkg.__version__))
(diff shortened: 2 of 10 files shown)


jspricke (approved):

bloom/logging.py
```diff
@@ -60,7 +55,7 @@
 
 _emoji_check_mark = "✅  "
 _emoji_cross_mark = "❌  "
-_is_mac_lion_or_greater = parse_version(mac_ver()[0]) >= parse_version('10.7.0')
+_is_mac = (mac_ver()[0] is not '')
```
jspricke, line 58:
```suggestion
_is_mac = (mac_ver()[0] != '')
```

bloom/commands/git/import_upstream.py
```diff
@@ -39,7 +39,7 @@
 import tarfile
 import tempfile
 
-from pkg_resources import parse_version
+from packaging.version import parse
```
nuclearsandwich, line 42: Being a bit of a python outsider, I've always disliked that a function's utility is sometimes obfuscated when the functions are imported outside their module.
Does using the alias-on-import feature make sense so that `parse` doesn't lose its context.
If using the alias-on-import feature is not seen as idiomatic then I could live with it as is, but I think it improves local legibility.

```suggestion
from packaging.version import parse as parse_version
```

bloom/commands/git/import_upstream.py
```diff
@@ -87,7 +87,7 @@ def version_check(version):
     info(fmt("The latest upstream tag in the release repository is '@!{0}@|'."
          .format(last_tag.replace('@', '@@'))))
     # Ensure the new version is greater than the last tag
-    if parse_version(version) < parse_version(last_tag_version):
+    if parse(version) < parse(last_tag_version):
```
nuclearsandwich, line 90: If either suggestion above is taken then this change would need to be reverted or modified.

bloom/commands/git/import_upstream.py
nuclearsandwich: Another alternative that is used in the other files is to import the entire version module which would afford the same context as the function would be invoked as `version.parse()`.

```suggestion
from packaging import version
```

bloom/generators/release.py
```diff
@@ -53,8 +53,8 @@
 
 try:
     import catkin_pkg
-    from pkg_resources import parse_version
-    if parse_version(catkin_pkg.__version__) < parse_version('0.3.8'):
+    from packaging import version
+    if version.parse(catkin_pkg.__version__) < version.parse('0.3.8'):
```
nuclearsandwich, line 57: @cottsay @wjwwood what do you think about removing these inline version checks entirely? bloom now requires catkin-pkg >= 0.4.3 according to `setup.py` and `stdeb.cfg` so this is just one more place where version dependencies have to be managed.

bloom/git.py
```diff
@@ -686,7 +687,9 @@ def get_last_tag_by_version(directory=None):
     versions = []
     for line in output.splitlines():
         tags.append(line.strip())
-        versions.append(parse_version(line.strip()))
+        ver = re.match("[0-9]+.[0-9]+.[0-9]+", line)
```
nuclearsandwich, line 690:
```suggestion
        ver = re.match(r"[0-9]+\.[0-9]+\.[0-9]+", line)
```

Since dots are regex metacharacters they need to be escaped for an exact match. Or is the intent that any character could separate these digits?

bloom/logging.py
```diff
@@ -60,7 +55,7 @@
 
 _emoji_check_mark = "✅  "
 _emoji_cross_mark = "❌  "
-_is_mac_lion_or_greater = parse_version(mac_ver()[0]) >= parse_version('10.7.0')
+_is_mac = (mac_ver()[0] != '')
```
nuclearsandwich, line 58: In general I'm fine with dropping "support" for Mac OS Lion+, let's make sure that's a full bump for the next release though.

+ Snow Leopard was the last great release of Mac OS anyway.

nuclearsandwich (changes requested):
Thanks @ijnek for this PR! and @jspricke for an earlier review!

While I'm not thrilled about the introduction of an additional dependency, it is one that is part of the extended suite of python packaging related dependencies and is widely packaged and available so it's not overly concerning.

This switch also resolves CI which is currently failing due to an apparent behavior change in how the deprecated pkg_resources function handles empty strings.

I have a couple of non-blocking suggestions but found one regex that needs feedback as I think it's incorrect as-is.

I also raised a question or two of general maintenance surfaced by looking at some changed areas which do not block merging this PR.

Once the regex is looked at again this should be quite close to ready.

bloom/commands/git/import_upstream.py
wjwwood: I would say the function name should just be something more specific than parse, but I think your first suggestion is just fine for now.

bloom/generators/release.py
wjwwood: I think that's probably fine, but I think historically I think this was in place to catch things like catkin_pkg is from pip and bloom is from apt or something like that, where the requirements can be circumvented. And this just a slightly nicer error than you'd get otherwise. Maybe we could extract this automatically some how so that you wouldn't need to update it manually.

bloom/git.py
wjwwood: This is likely what was intended.

bloom/commands/git/import_upstream.py
ijnek: It looks like I've avoided using ``from packaging import version``, because the parameter for the ``version_check`` method was also ``version`` and there was a conflict.

But I can go with parse_version.

ijnek:
Across all files, I've gone with ``from packaging.version import parse as parse_version``, because the ``parse`` isn't descriptive enough, and ``version.parse`` causes a lot of conflicts with local variables and parameters named ``version``.

bloom/git.py
nuclearsandwich: Resolved in cd2031799a3e00296d70a47f0440326c67ea64d6

nuclearsandwich (approved):
Thanks for iterating @ijnek and for this excellent improvement. Among other things I believe this addresses the issue that is affecting CI for all Bloom PRs right now.

PR #698, opened 2023-03-10
Add debian dependency to stdeb.cfg for new packaging dependency.

nuclearsandwich:
This was introduced in #693 and I realized that it needed the stdeb dependency while I was on a walk but forgot to add that before re-reviewing and merging that PR.

This package is available for Python 2 in [Buster](https://packages.debian.org/buster/python3-packaging) and [Bionic](https://packages.ubuntu.com/bionic/python-packaging).

As well as for Python 3 in [Buster](https://packages.debian.org/buster/python3-packaging) and [Bionic](https://packages.ubuntu.com/bionic/python3-packaging) and later distributions.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,8 +1,8 @@
 [DEFAULT]
 ; release with a high debinc to avoid conflict with upstream debian of the same release version
 Debian-Version: 100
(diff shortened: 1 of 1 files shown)


ijnek (approved):
Thanks for this. Just fyi - you've got a typo in the issue linked . it should be #693

nuclearsandwich:
> Just fyi - you've got a typo in the issue linked . it should be #693

Good catch! Fixed in both the PR and the commit body.

cottsay (approved):

PR #702, opened 2023-05-26
Introduce notion of "experimental" build types.

nuclearsandwich:
When per build type templates were introduced for ROS 2 support, little thought was given to how the number of build types might expand in the future. I think that there's good fodder for a rethink of the build type handling in general, but in the meantime I would like to introduce this concept and a small warning to signify the possibility that these new build types are outside the focus of the core infrastructure team.

The initial criteria for a build type being "experimental" is that it's novel.
Criteria for graduating from an experimental build type to normal one is not currently settled.


nuclearsandwich:
I never had the bandwidth to get this change across the finish line and I no longer have the enthusiasm for it that I once did. In the long term, I think that bloom should take inspiration (if not actual code) from the colcon approach and make providing build templates an extension point so they can come from other packages outside the core bloom repository.

PR #703, opened 2023-06-12
Generate -runtime packages with --runtime-pkg

jspricke:
This allows separating runtime and development packages to cut the dependency
tree. Note that all packages would need to be created with the --runtime-pkg
option as there is no logic to store which packages has one and which one does
not. Dependencies are separated by exec_depend/build_export_depend.

The package split currently only moves the include directory as the cmake,
pkg-confg and package.xml files could be used by runtime scripts.
Also the package names are -runtime/ instead of /-dev so users get everything
in case they install the old package name.

In case you want to use this on packages.ros.org I would recommend to run
bloom-generate as part of the package build process.

The second patch adds a similar logic for test dependencies by using
test_depend only for runtime packages. Test dependencies that need development
files would need to use build_depend. This is a light redefinition of the tag
but allows to cut the dependency tree even further.


ros-discourse:

https://discourse.ros.org/t/package-separation-for-building-and-running-ros-applications/35527/5

bloom/generators/debian/templates/cmake/rules.em
```diff
@@ -59,11 +59,24 @@ override_dh_shlibdeps:
 	# in the install tree and source it.  It will set things like
 	# CMAKE_PREFIX_PATH, PKG_CONFIG_PATH, and PYTHONPATH.
 	if [ -f "@(InstallationPrefix)/setup.sh" ]; then . "@(InstallationPrefix)/setup.sh"; fi && \
-	dh_shlibdeps -l$(CURDIR)/debian/@(Package)/@(InstallationPrefix)/lib/:$(CURDIR)/debian/@(Package)/@(InstallationPrefix)/lib/${DEB_HOST_MULTIARCH}
+@[if RuntimePackage]
+	dh_shlibdeps -l$(CURDIR)/debian/@(Package)-runtime@(InstallationPrefix)/lib/:$(CURDIR)/debian/@(Package)-runtime@(InstallationPrefix)/lib/${DEB_HOST_MULTIARCH}
+@[else]
+	dh_shlibdeps -l$(CURDIR)/debian/@(Package)-runtime@(InstallationPrefix)/lib/:$(CURDIR)/debian/@(Package)-runtime@(InstallationPrefix)/lib/${DEB_HOST_MULTIARCH}
```
peci1, line 65: I guess the -runtime should not be here in this brach...

jspricke: fixed, thanks!

PR #704, opened 2023-06-12
Use ros- prefix for bloom-generate debian

jspricke:
Copied from bloom/generators/rosdebian.py.


nuclearsandwich:
Can you elaborate on the utility of this change?
The debian generator is intended to be usable for non-ROS debian packages and so the change as is does not make sense to me.

jspricke:
Fine with me.

PR #705, opened 2023-06-15
Debian generator support for architecture_independent flag

frankkusters:
The `Architecture` field is set to `all` if a package has `metapackage` or `architecture_independent` in the exports in its `package.xml`.

This is analogous to the support in the RPM generator (https://github.com/ros-infrastructure/bloom/pull/270).


frankkusters:
Is there any action I should be taking to move forward on merging this PR?

nim65s:
ftr, I tried this, and ros buildfarm still use eg. `-DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu`

Issue #697, opened 2023-03-06
Generation of copyright files


ct2034:
## Background
Debian packages need copyright files according to this format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/. Currently, copyright files are created by this template [https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/ament_python/copyright.em](https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/ament_python/copyright.em). Which is producing copyright files that are not strictly following the format.
This is partly because not all the required information is available in current `package.xml`s. To add the needed information @ralph-lange created https://github.com/ros-infrastructure/rep/pull/347.
### ros_license_linter
The linter https://github.com/boschresearch/ros_license_linter has two intended functionalities:
- checking if the license declaration in the `package.xml` is correct, by scanning the code for existing licenses and comparing that to the declared licenses.
- assembling the information for the copyright files
*But currently, only the first functionality is implemented.*
In a second step, it makes sense to integrate the linter into ament_lint and to make it available as github action.
### Why should we have the functionality in a separate package?
There is quite some overlap between the two functionalities of license checking and copyright file creation. So, integrating the functionality to generate copyright files into bloom would create a lot of code duplication when also integrating it into ament_lint.
## My Question
With this issue, I would like to ask you @wjwwood, whether you generally agree with this. Because I would then start to implement the PR that would add https://github.com/boschresearch/ros_license_linter as a dependency of bloom and improve the generation of copyright files to match the required format.
## Further steps
- In a second step, it would also be desirable to have every package checked by the linter on a bloom release to verify that there are no undeclared licenses in the code. But I think that is a separate discussion.
- I am generally open to moving the license linter into the `ros-infrastructure` org and to maybe renaming it to something more suitable.

ipa-cmh:
I think this is a very good idea! Also looking very much forward to having correct copyright files!

wjwwood:
This is a question for @nuclearsandwich and @cottsay. Having a dependency on that tool might be acceptable, but they would know better.

In general, I support the idea, it's something we always wanted, but never had the resources to make it happen.

ct2034:
Thanks for the support @wjwwood. Sorry, I assumed you were the maintainer.

alltimaterobotics:
That really make it easy to have better overview of package copyrights . And that's vital for all of us, having commercial use of ROS in mind.
[...]

PR #709, opened 2023-09-27
Generation of copyright files

ct2034:
This is my implementation of #697.
It is still work in progess

### DIFF
--- a/bloom/generators/debian/templates/ament_cmake/copyright.em
+++ b/bloom/generators/debian/templates/ament_cmake/copyright.em
@@ -1,11 +1,1 @@
-Format: Bloom subset of https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
-Upstream-Name: @(Name)
-@[if BugTracker]Upstream-Contact: @(BugTracker)@\n@[end if]@
-@[if Source]Source: @(Source)@\n@[end if]@
-@[for License, Text in Licenses]@
-
-Files: See file headers in repository for details
-Copyright: See package copyright in source code for details
-License: @(License)
- @(Text)
-@[end for]@
+@(Copyright_file_content)
--- a/bloom/generators/debian/generator.py
+++ b/bloom/generators/debian/generator.py
@@ -108,6 +108,12 @@ except ImportError:
     debug(traceback.format_exc())
     error("empy was not detected, please install it.", exit=True)
 
+try:
+    from ros_license_toolkit.package import get_packages_in_path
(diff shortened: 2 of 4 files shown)


bloom/generators/debian/templates/ament_cmake/copyright.em
```diff
@@ -1,11 +1 @@
...
-@[if BugTracker]Upstream-Contact: @(BugTracker)@\n@[end if]@
-@[if Source]Source: @(Source)@\n@[end if]@
-@[for License, Text in Licenses]@
-
-Files: See file headers in repository for details
-Copyright: See package copyright in source code for details
-License: @(License)
- @(Text)
-@[end for]@
+@(Copyright_file_content)
```
tfoote, line 1: The capitalization of this variable name is unexpected `copyright_file_content` would be more consistent, here and in 4 other places.

ct2034: Honestly, I think for consistency, it should be capitalized. If you look at the other variables, let alone in this snippet: `Text`, `License`, ...

tfoote: Well that's unfortunate. I guess stick with what's local then.

ct2034:
At ROSCon I talked to @nuclearsandwich and @clalancette. This is my summary what we agreed to do:
Yes, the [REP change](https://github.com/ros-infrastructure/rep/pull/347) was accepted in the [TSC](https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-06-15/32038).
I would like to merge [this](https://github.com/ros-infrastructure/bloom/pull/709) integrating [ros_license_linter](https://github.com/boschresearch/ros_license_linter).
But there are problems because:
a) This must increment the package.xml schema version.
b) @clalancette wants to do other changes then, too.
So, the solution was to change the tooling such that it
a) differentiates between package.xml versions (not requiring the new attributes introduced by [REP change](https://github.com/ros-infrastructure/rep/pull/347)),
b) And it __never__ cause bloom to fail.
These changes are now implemented on our side.

ct2034:
As an update: We think, we can get rid of the dependency on spdx-tools (https://github.com/boschresearch/ros_license_toolkit/blob/63c7406866a5f81eee1457922a6b17fb6b0fd2df/pyproject.toml#L31) but we will probably have to keep the dependency on scancode-toolkit. What are the next steps, then?

PR #736, opened 2024-06-18
Update for Python 3

jspricke:


nuclearsandwich (commented):
Thanks for the update. I think it's a welcome change, but I'm not actually certain that we've dropped support for Python 2 officially yet in bloom.

#733 claims to still support 2.7.

I am happy with this PR as something we merge during a 0.13.0 window dropping Python2 support and will aim to have such a release out before the end of this month.

PR #745, opened 2025-06-24
Replace vcstools with vcstool

esteve:
This PR is based on @cottsay work at https://github.com/ros-infrastructure/bloom/tree/cottsay%2Fvcstool, I've only rebased it and replaced a `vcstools` import.

### DIFF
--- a/bloom/commands/export_upstream.py
+++ b/bloom/commands/export_upstream.py
@@ -58,12 +58,7 @@ from bloom.util import change_directory
 from bloom.util import handle_global_arguments
 from bloom.util import temporary_directory
 
-try:
-    from vcstools.vcs_abstraction import get_vcs_client
-except ImportError:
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
Thanks @esteve, this branch was still work-in-progress when last touched and was set aside in part because `vcstool` itself was not being actively maintained. We have a current Google Summer of Code project from @leander-dsouza to bring an updated vcstool to PyPI and the ROS apt repositories under a new name since we're unable to reach the current maintainer on PyPI and gain control of the project. One of the stretch goals for that project is to get bloom switched over to using it, but if you have interest in doing some of that work in parallel to that do feel free to add commits to this branch. For our part, we'll likely only start this work once https://github.com/ros-infrastructure/vcs2l is released on PyPI and the ROS bootstrap repository in a usable state.

PR #746, opened 2025-06-25
Fix no remote check.

nuclearsandwich:
This check has likely never worked as intended. I'm not aware of any case where `len()` returns a negative number. The described error seems like the intent was to check for an empty list of remotes not a somehow negative number of remotes.

When updating this check, there are several test failures which  i have not triaged. It's possible that now-expected behavior is incompatible with this remote check or that our test fixtures need to be updated so they don't cause this error.


nuclearsandwich:
FYI @wjwwood in case there was once a valid way for `len(remotes) < 0` to be true.

wjwwood:
I can't think of one, likely a bug that prevented us from catching this condition.

cottsay (approved):

cottsay:
Hmm, this breaks a bunch of tests. Converting to draft until someone has time to come up with a plan.

PR #747, opened 2025-06-30
Replace pkg_resources usage with importlib.

nuclearsandwich:
Cleaning up bloom for a release and getting pelted with these deprecation warnings. Apparently pkg_resources will be removed in setuptools 81:

> The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.

---

This might fail CI as I'm using the latest stdlib importlib on Arch and may need to up the minimum python version against the backports. CI will tell.


cottsay (approved):
Thanks.

I'll say that elsewhere we're using a try/except to attempt use of the standard library import and falling back to the standalone package rather than using conditionals based on the version, but I don't see a particular reason that this wouldn't work as well.

nuclearsandwich:
> I'll say that elsewhere we're using a try/except to attempt use of the standard library import and falling back to the standalone package rather than using conditionals based on the version

In some cases, even when the stdlib library is available (e.g. 3.7+) the specific API functions that we're using were only added later. So library presence is not a sufficient condition to check and we can't rely on the import error alone. To use try-exceptw we'd have to catch the argument error down the line which is pretty nasty.

Rather than think about whether I could use try-import-except or had to check the stdlib version I just check the stdlib version everywhere.

Issue #750, opened 2025-09-04
Only do a shallow clone of rosdistro


azeey:
Bloom currently pulls the entire rosdistro repo everytime one makes a release. With how active that repo is, the `git pull` in the following snippet downloads about 122 MiB according to `git`. 
https://github.com/ros-infrastructure/bloom/blob/6199b2a76f50f120a398c48c44b1e1ffd6c5f38f/bloom/commands/release.py#L765-L767

Using a shallow clone or `git fetch --depth 1 && git merge FETCH_HEAD` downloads about 562 KiB. Unless there a specific reason for fetching the entire history, I think this would speed up bloom a bit.

PR #755, opened 2026-01-22
Speed up rosdistro cloning with blobless partial clones

azeey:
Replaces the manual `mkdir`, `git init`, and `git pull` sequence used for cloning the rosdistro repository with an optimized `git clone`.

I had brought this up before in https://github.com/ros-infrastructure/bloom/issues/750, but shallow clones weren't the right approach.

Key changes:
- Uses [`--filter=blob:none`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---filterfilter-spec) (blobless clone) to drastically reduce the download size by omitting historical file contents. **Reference:** [GitHub Blog - Get up to speed with partial clone and shallow clone](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/)
- Ensures that OAuth tokens are not persisted in the local git config.
- Includes a fallback to standard clones for Git versions that do not support partial clones (pre-2.20).

This change improves both performance and security during the pull request generation phase of the release process.

I've tested successfully on a couple of releases: 
- https://github.com/ros/rosdistro/pull/49559
- https://github.com/ros/rosdistro/pull/49560

Generated-By: Gemini 2.5 Pro


mjcarroll (approved):
Wow! This is great!

cottsay:
@azeey @mjcarroll, I made modifications to this PR to reduce the number of changes but keep the same "spirit". PTAL and provide feedback.

bloom/commands/release.py
```diff
@@ -770,9 +770,13 @@ def _my_run(cmd, msg=None):
...
                     return
-                _my_run('git checkout -b {new_branch}'.format(**locals()))
-                _my_run("git pull {rosdistro_url} {base_info[branch]}".format(**locals()),
-                        "Pulling latest rosdistro branch")
+                # Use partial fetch if available to speed up cloning large rosdistro repo
+                try:
+                    _my_run('git fetch --filter=blob:none {rosdistro_url} HEAD:{new_branch} -n'.format(**locals()))
+                except subprocess.CalledProcessError:
+                    info('Partial fetch failed, falling back to regular fetch...')
+                    _my_run('git fetch {rosdistro_url} HEAD:{new_branch} -n'.format(**locals()))
```
nuclearsandwich:
```suggestion
                    _my_run('git fetch {rosdistro_url} {base_info[branch]}:{new_branch} -n'.format(**locals()))
```

Unless it was intentional to always fetch the default remote HEAD `base_info[branch]` should probably still be used as it was in deleted line 774.

bloom/commands/release.py
```diff
@@ -770,9 +770,13 @@ def _my_run(cmd, msg=None):
                 if interactive and not maybe_continue():
                     warning("Skipping the pull request...")
                     return
-                _my_run('git checkout -b {new_branch}'.format(**locals()))
-                _my_run("git pull {rosdistro_url} {base_info[branch]}".format(**locals()),
-                        "Pulling latest rosdistro branch")
+                # Use partial fetch if available to speed up cloning large rosdistro repo
+                try:
+                    _my_run('git fetch --filter=blob:none {rosdistro_url} HEAD:{new_branch} -n'.format(**locals()))
```
nuclearsandwich:
```suggestion
                    _my_run('git fetch --filter=blob:none {rosdistro_url} {base_info[branch}:{new_branch} -n'.format(**locals()))
```

Unless it was intentional to always fetch the default remote HEAD `base_info[branch]` should probably still be used as it was in deleted line 774.

nuclearsandwich: Reading https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/ I would have expected a "treeless" fetch be better than a "blobless" fetch.

When actually running comparisons, although a treeless fetch saved hundreds of thousands of objects and about 6MiB it was on average 0.6 - 0.9 seconds slower to actually clone.


```
# treeless
Receiving objects: 100% (65832/65832), 26.76 MiB | 16.16 MiB/s, done.
# blobless
Receiving objects: 100% (181253/181253), 33.91 MiB | 15.15 MiB/s, done.
```

I could see this being an optimization issue on the server side and since we're trying to cut down on wall time blobless seems to be the correct path (although it would be good for others to corroborate my local samplesize=3 tests) and I wonder if treeless is still the answer for people with bandwidth constraints that would wash out the server-side optimization for blobless.

cottsay: 7c5e37f

bloom/commands/release.py
cottsay: 7c5e37f

cottsay:
~~I'm not 100% convinced that this actually saves us much overall - even though the fetch side of this is now lightning fast, pushing the results to the fork seems to take a lot longer.~~

**EDIT:** Nix that, the push is taking a long time on both. This is definitely an improvement.

PR #756, opened 2026-01-22
Optimize PR branch calculation 

azeey:
This uses `git ls-remote` instead of hitting the Github APIs for getting
a list of branches on the fork repo. This is faster and avoids Github's
rate limiting.

Generated-By: Gemini 2.5 Pro


cottsay (approved):
This looks good, thanks. I like that this limits the use of the GitHub token to only the creation of the pull request.

PR #763, opened 2026-04-13
Add RPM templates for 'cargo' build type

cottsay:
Only `cargo` build type, and only `rpm` templates at this stage.


bloom/generators/rpm/templates/cargo/template.spec.em
```diff
@@ -0,0 +1,67 @@
...
+@[if Supplements]@\n%if 0%{?with_weak_deps}
+@[for p in Supplements]Supplements:    @p@\n@[end for]@
+%endif@\n@[end if]@
+
+%description
+@(Description)
+
+%prep
+%autosetup -p1
+%cargo_prep -v /usr/share/cargo/registry
```
Blast545: Shouldn't this line use `%{cargo_registry}`?

bloom/generators/rpm/templates/cargo/template.spec.em
```diff
@@ -0,0 +1,67 @@
...
+pallet-patcher --output-format=toml Cargo.toml %{cargo_registry} > pallet-patcher.toml
+
+%build
+# In case we're installing to a non-standard location, look for a setup.sh
+# in the install tree and source it.  It will set things like
+# CMAKE_PREFIX_PATH, PKG_CONFIG_PATH, and PYTHONPATH.
+if [ -f "@(InstallationPrefix)/setup.sh" ]; then . "@(InstallationPrefix)/setup.sh"; fi
+%cargo_build -- --config pallet-patcher.toml
```
Blast545, line 46: Should we pass to `cargo_build` somewhere `cargo_registry`? Or is it a name that will be taken automatically?

Blast545 (commented):
I'm not quite familiar with bloom, so I made a first pass review by comparing this file to https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/rpm/templates/ament_python/template.spec.em and making sure the flow of the process sounds reasonable. For now LGTM after clarification questions.

Is there a "simple" way to test locally if this works? 👀

bloom/generators/rpm/templates/cargo/template.spec.em
cottsay: Hmm, this template was created when I thought it was going to be possible to vendor all of the dependencies "as usual", and I think it's clear that isn't going to work.

The line is somewhat correct given my original intent. We changed `cargo_registry` to the ROS install path at the top of this file, but this is the one place where we actually wanted to reference the system's crate stash. Then we're using pallet-patcher to search the ROS registry later.

For the sake of making progress on this template, let's forego source replacement entirely for now and let `cargo` download dependencies from `crates.io` live. When we have a viable solution for vendoring those dependencies (possibly via bloom itself), we can update it here to use that per-package vendor directory.

As of now, I don't think we have any plans to use the system's crate registry.


```suggestion
%cargo_prep -N
```

bloom/generators/rpm/templates/cargo/template.spec.em
cottsay: The way RPM macros work, they're already implicitly referencing the `cargo_registry` macro. In a sense they work more similarly to CMake macros than CMake functions, and can see other macros that are defined when they are invoked.

bloom/generators/rpm/templates/cargo/template.spec.em
Blast545: > Hmm, this template was created when I thought it was going to be possible to vendor all of the dependencies "as usual", and I think it's clear that isn't going to work.

Why that's not possible? I'm assuming we can vendor the dependencies of a package OR use local registry to get deps from there with pallet-patcher OR a mix between the two. 

If you mean setting `%cargo_prep -N` for the sake of testing the rpm for now, I think that sounds reasonable

cottsay: That wasn't a very clear thought, let me explain what I meant.

By "as usual", I was referring to the way platforms are doing this. All dependencies are packaged using the system package manager and nothing comes from external package managers (like pip or crates.io), and very little or nothing is bundled along with each package.

I think that in the end, we'll end up with something more like a bundling approach, but either way I don't see us using the system's registry at `/usr/share/cargo/registry` because it really can't provide a "complete" solution for us. When I wrote this code, I was hoping that it would.

bloom/generators/rpm/generator.py
```diff
@@ -273,17 +273,14 @@ def generate_substitutions_from_package(
...
-    elif build_type == 'ament_python':
-        pass
-    else:
+    if build_type not in (
+        'cargo',
+        'catkin',
+        'cmake',
+        'meson',
+        'ament_cmake',
+        'ament_python',
```
cottsay: Alphabetize

```suggestion
        'ament_cmake',
        'ament_python',
        'cargo',
        'catkin',
        'cmake',
        'meson',
```

PR #764, opened 2026-04-15
Enable dynamic RPM generator

cottsay:
The legacy RPM generator will now be disabled for RHEL < 10 and Fedora < 43. The dynamic RPM generator runs very quickly and will be enabled if Fedora or RHEL are supported in the target ROS distro.

This will result in a change to the "action list", and the first release operation for every release repository will prompt the developer to update to the new default action list. The developer can accept the default at the prompt (yes) and their action list will automatically move to the new list.

### DIFF
--- a/bloom/generators/dynrpm/generator.py
+++ b/bloom/generators/dynrpm/generator.py
@@ -371,10 +371,21 @@ class DynRpmGenerator(BloomGenerator):
                  "even if not in current upstream")
         add('--install-prefix', default=None,
             help="overrides the default installation prefix (/usr)")
+        add('--only-os-names', nargs='*', required=False,
+            help="skip generation if rosdistro doesn't have any of "
+                 "these platforms listed")
 
     def handle_arguments(self, args):
         self.interactive = args.interactive
         self.rpm_inc = args.rpm_inc
+        if args.only_os_names:
+            index = rosdistro.get_index(rosdistro.get_index_url())
+            distribution_file = rosdistro.get_distribution_file(index, self.rosdistro)
+            if not set(args.only_os_names).intersection(distribution_file.release_platforms):
+                warning("No platforms defined for given OS filter in release file for the '{0}' distro."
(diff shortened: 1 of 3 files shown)


bloom/generators/dynrpm/generator.py
```diff
@@ -371,10 +371,21 @@ def prepare_arguments(self, parser):
                  "even if not in current upstream")
         add('--install-prefix', default=None,
             help="overrides the default installation prefix (/usr)")
+        add('--only-os-names', nargs='*', required=False,
+            help="skip generation if rosdistro doesn't have any of "
+                 "these platforms listed")
```
claraberendsen: I'm not sure the naming of the arg is clear enough for someone without the helper context.
Maybe something like `--require-os` is more self explanatory.

claraberendsen (approved):
Overall looks good to me. Left a small non blocking comment on naming. 

From reading the code I understand that different from the description of the PR, the dynrpm will be enable RHEL 10+ and Fedora 43+ disabling the legacy rpm builder correct?

That means that RHEL <10 and Fedora <43 will stay on the legacy rpm.

cottsay:
> From reading the code I understand that different from the description of the PR, the dynrpm will be enable RHEL 10+ and Fedora 43+ disabling the legacy rpm builder correct?

The dynamic RPM generator will be enabled wherever there's a ROS distribution with RHEL and/or Fedora in the distribution.yaml.

The legacy RPM generator will only be enabled if there is a RHEL version prior to 10 or a Fedora version prior to 43.

bloom/generators/dynrpm/generator.py
cottsay: sure, fine with me. The generator is so fast that I'm tempted to run it unconditionally, but this is probably the "right" thing to do.

cottsay: 9544a04

claraberendsen:
LGTM! Thanks for the changes :heart:

PR #774, opened 2026-05-05
Add Debian template for 'cargo' build type

Blast545:
Adds a Debian/Ubuntu packaging template for the `cargo` build_type, mirroring the scope of #763 (RPM cargo template).                                                                                                         

  - Embeds a dependency SBOM in shipped binaries via `cargo auditable` (compressed JSON in a `.dep-v0` ELF section)
  - v1 lets cargo fetch from crates.io at build time.


bloom/generators/debian/templates/cargo/rules.em
```diff
@@ -0,0 +1,83 @@
...
+export DH_VERBOSE=1
+# TODO: remove the LDFLAGS override.  It's here to avoid esoteric problems
+# of this sort:
+#  https://code.ros.org/trac/ros/ticket/2977
+#  https://code.ros.org/trac/ros/ticket/3842
+export LDFLAGS=
+export PKG_CONFIG_PATH=@(InstallationPrefix)/lib/pkgconfig
+# Explicitly enable -DNDEBUG, see:
+# 	https://github.com/ros-infrastructure/bloom/issues/327
+export DEB_CXXFLAGS_MAINT_APPEND=-DNDEBUG
```
Blast545, line 19: I was wondering, do we still need those for `cargo`? I was reading some of those and they seem related to C++ optimizations.

bloom/generators/debian/templates/cargo/control.em
```diff
@@ -0,0 +1,14 @@
+Source: @(Package)
+Section: misc
+Priority: optional
+Maintainer: @(Maintainer)
+Build-Depends: debhelper (>= @(debhelper_version).0.0), @(', '.join(sorted(set(BuildDepends) | {'cargo', 'cargo-auditable', 'dh-cargo', 'pallet-patcher', 'rustc'})))
```
Blast545: cargo-auditable is only needed if we agree that https://discourse.ubuntu.com/t/supply-chain-accountability-coming-to-a-rust-package-near-you/81214 it's a good solution moving forward to deal with SBOM traceability of rust crates generated on the buildfarm.

Before realizing this, I just generated a .txt file with all the packages used to compile the project from the target folder. I think cargo-auditable it's more robust and will help the crates the generated in the buildfarm to co-exist with other tools in the ecosystem.

bloom/generators/debian/templates/cargo/rules.em
```diff
@@ -0,0 +1,81 @@
...
+DEB_HOST_GNU_TYPE ?= $(shell dpkg-architecture -qDEB_HOST_GNU_TYPE)
+
+%:
+	dh $@@ -v --buildsystem=cargo --builddirectory=.obj-$(DEB_HOST_GNU_TYPE)
+
+override_dh_auto_configure:
+	# dh-cargo's configure step expects these two files to exist, even when
+	# we aren't vendoring dependencies. Touching an empty cargo-checksum.json
+	# and Cargo.lock is the canonical workaround (see Debian wiki: Rust Packaging).
+	touch debian/cargo-checksum.json Cargo.lock
```
Blast545, line 30: I tried using the Cargo.lock from the project and include it in the debian generated. Since that cargo.lock also contains the resolution for platforms different than the target platform, I wonder how useful is that and if it might be misleading for users consuming the package. 

So, I'm leaning towards not including those.

bloom/generators/debian/templates/cargo/control.em
```diff
@@ -0,0 +1,14 @@
...
+Homepage: @(Homepage)
+Standards-Version: 4.6.0
```
cottsay, line 7: Note that this is the same version as the `meson` templates, but is newer than our existing `cmake` and `ament_python` templates, which are at 3.9.2.

No action needed, just wanted to point that out.

bloom/generators/debian/generator.py
```diff
@@ -393,15 +393,14 @@ def generate_substitutions_from_package(
...
-    elif build_type == 'ament_cmake':
-        pass
-    elif build_type == 'ament_python':
+    if build_type in (
+        'cargo',
+        'catkin',
+        'cmake',
+        'meson',
+        'ament_cmake',
+        'ament_python',
```
cottsay: I know they weren't in order before, but since we're touching these lines anyway, let's alphabetize them.

```suggestion
        'ament_cmake',
        'ament_python',
        'cargo',
        'catkin',
        'cmake',
        'meson',
```

bloom/generators/debian/templates/cargo/rules.em
cottsay: I believe it's possible to build C/C++ code using Cargo using a `build.rs`, but I'm not sure if debian is able to plumb through the "standard" way we express what compile flags we'd like used into Cargo.

To be determined. Maybe we should leave them in for now since they shouldn't cause any harm.

bloom/generators/debian/templates/cargo/rules.em
```diff
@@ -0,0 +1,81 @@
...
+	# In case we're installing to a non-standard location, look for a setup.sh
+	# in the install tree and source it.  It will set things like
+	# CMAKE_PREFIX_PATH, PKG_CONFIG_PATH, and PYTHONPATH.
+	if [ -f "@(InstallationPrefix)/setup.sh" ]; then . "@(InstallationPrefix)/setup.sh"; fi && \
+	dh_auto_configure
+	# Generate a pallet-patcher config so that, once we have a viable
+	# vendoring solution, we can point cargo to local deps
+	pallet-patcher --output-format=toml Cargo.toml > pallet-patcher.toml
```
cottsay: We need to search our prefix implicitly, right?

```suggestion
	pallet-patcher --output-format=toml Cargo.toml @(InstallationPrefix)/share/cargo/registry > pallet-patcher.toml
```

bloom/generators/debian/templates/cargo/rules.em
cottsay: I don't think that downstream build processes will implicitly use them at all, right? If they do no harm, I'm inclined to leave the default behavior alone.

bloom/generators/debian/templates/cargo/rules.em
```diff
@@ -0,0 +1,81 @@
...
+
+override_dh_auto_test:
+	# In case we're installing to a non-standard location, look for a setup.sh
+	# in the install tree and source it.  It will set things like
+	# CMAKE_PREFIX_PATH, PKG_CONFIG_PATH, and PYTHONPATH.
+	echo -- Running tests. Even if one of them fails the build is not canceled.
+	if [ -f "@(InstallationPrefix)/setup.sh" ]; then . "@(InstallationPrefix)/setup.sh"; fi
+	cargo test || true
+
+override_dh_auto_install:
```
cottsay, line 66: By overriding `dh-cargo`'s install we're missing out on this whole section, which is responsible for installing the unpacked crate to `<prefix>/share/cargo/registry/`

https://github.com/deepin-community/dh-cargo/blob/00d6ee3b093ef085d02302bc9fc1e4bdc821fc14/cargo.pm#L177-L186

(sorry I couldn't figure out how to link to dh-cargo's actual upstream source code, but this fork/mirror seems to work)

cottsay: Unfortunately the output path is hard-coded in that snippet, so I wonder if the best thing to do is maybe to let the install happen to `/usr/share/cargo/registry` and then blindly move the results (if any) to `@(InstallationPrefix)/share/cargo/registry`.

Blast545:
Do not review, iterating on this one. After I recalled that crates are separated debians/libs, see: https://github.com/deepin-community/dh-cargo/blob/00d6ee3b093ef085d02302bc9fc1e4bdc821fc14/cargo.pm#L177-L186, I created a more complete example for testing this PR and it's not completely working yet.

Blast545:
Ok, I tested this with: https://github.com/Blast545/testing_colcon_ros_buildfarm_cargo. There's a lot of llm generated doc, but the interesting parts of the setup that are worth checking:

Some dependencies we need in the buildfarm to build cargo stuff: https://github.com/Blast545/testing_colcon_ros_buildfarm_cargo/tree/main#patches-applied-to-the-venv-install
(Including how do we want to install pallet-patcher, which is not available via apt right now)

rosdistro/distribution.yaml: https://github.com/Blast545/testing_colcon_ros_buildfarm_cargo/blob/main/configs/rosdistro/rolling/distribution.yaml

The rust crates under the test:
https://github.com/Blast545/ros_rust_lib_demo/blob/main/package.xml
https://github.com/Blast545/ros_rust_app_demo/blob/main/package.xml

bloom/generators/debian/templates/cargo/rules.em
Blast545: I think there's no simple way to rely on their code to do what they do with our custom naming/path needs, so I think we can just try to go on our own route. At least for the proof of concept

bloom/generators/debian/templates/cargo/rules.em
```diff
@@ -0,0 +1,101 @@
...
+	#     embeds a compressed JSON SBOM into each binary's .dep-v0 ELF
+	#     section. Tools that read it: cargo audit bin, trivy, grype, syft,
+	#     osv-scanner, rust-audit-info. --no-track suppresses .crates.toml.
+	set -e ; \
+	if [ -f "@(InstallationPrefix)/setup.sh" ]; then . "@(InstallationPrefix)/setup.sh"; fi ; \
+	META=$$(cargo metadata --no-deps --format-version 1) ; \
+	CRATE_NAME=$$(printf '%s' "$$META" | python3 -c "import sys,json;print(json.load(sys.stdin)['packages'][0]['name'])") ; \
+	CRATE_VER=$$(printf '%s' "$$META" | python3 -c "import sys,json;print(json.load(sys.stdin)['packages'][0]['version'])") ; \
+	HAS_LIB=$$(printf '%s' "$$META" | python3 -c "import sys,json;p=json.load(sys.stdin)['packages'][0];L={'lib','rlib','dylib','cdylib','staticlib','proc-macro'};print(int(any(k in L for t in p['targets'] for k in t['kind'])))") ; \
+	HAS_BIN=$$(printf '%s' "$$META" | python3 -c "import sys,json;p=json.load(sys.stdin)['packages'][0];print(int(any('bin' in t['kind'] for t in p['targets'])))") ; \
```
Blast545, line 85: I'm not sure if there's a simpler way to detect if a specific package has libraries or is it only binary.

PR #777, opened 2026-06-26
Drop 'path' from Cargo dependencies during rebase

cottsay:
The Bloom rebasing process involves trimming the repository contents to only include the package being released. Cargo manifests can reference other packages in the repository using a 'path' value as part of the dependency spec, which is no longer valid.

Remove all 'path' values from dependencies and add a placeholder 'version' value if one is not already present such that the manifest is still valid.

---

Eventually, we'll want this normalization process to be more involved so that we can de-workspace the cargo packages as well. This is done as part of `cargo package` and `cargo read-manifest`.

I'm not 100% certain that this is the right place to do this. Maybe it should be an implicit patch, or should be moved to `rosrelease.py`, or should maybe be conditional on detecting a package of type `cargo` or `ament_cargo`. I'm open to suggestions.

### DIFF
--- a/bloom/generators/release.py
+++ b/bloom/generators/release.py
@@ -195,7 +195,7 @@ Please checkout the release branch and then create a tag manually with:""")
 
         for spec in (
             spec for category in (
-                'dependenices',
+                'dependencies',
                 'dev-dependencies',
                 'build-dependencies',
             ) for spec in data.get('dependencies', {}).values()
--- a/bloom/generators/release.py
+++ b/bloom/generators/release.py
@@ -32,7 +32,16 @@
 
 from __future__ import print_function
 
+import os
 import traceback
+
+import tomli_w
+
+try:
+    import tomllib
+except ImportError:
+    # For python < 3.11
+    import tomli as tomllib
 
 from bloom.generators import BloomGenerator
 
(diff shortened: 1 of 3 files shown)


bloom/generators/release.py
```diff
@@ -168,3 +180,31 @@ def detect_branches(self):
...
+        if not os.path.isfile(manifest):
+            return
+
+        print('Normalizing Cargo.toml')
+        with open(manifest, 'rb') as f:
+            data = tomllib.load(f)
+
+        for spec in (
+            spec for category in (
+                'dependenices',
```
Blast545:
```suggestion
                'dependencies',
```

bloom/generators/release.py
```diff
@@ -126,6 +135,9 @@ def post_rebase(self, destination):
         # Execute trim
         if trim_d in ['', '.']:
             return
+        ret = self._normalize_cargo_manifest(trim_d)
+        if ret:
```
Blast545, line 139: I think we could make a comparison with `package.get_build_type()` and avoid calling this function for non-cargo types. But I also think it would be super weird for non cargo types to have a Cargo.toml, so I think this is fine for now to unblock us.

Blast545 (approved):
Small typo, otherwise LGTM

PR #780, opened 2026-09-02
Fix dynrpm handling of '<replaces>'

cottsay:
The 'Obsoletes:' tag in RPM spec files cannot carry virtual package names like the others. Unfortunately, this is the only tag where we'll have to reference an actual package name in `bloom-deps.py`.


odra:
Looks good to me, ran all tests as well.

PR #781, opened 2026-09-02
Only remove sibling paths from cargo manifests

cottsay:
We can leave direct references to crates in subdirectories.

This code hasn't been released yet so it's still safe to make critical changes to it.

Assisted-by: Gemini 3.5 Flash

### DIFF
--- a/test/unit_tests/test_generators/test_release.py
+++ b/test/unit_tests/test_generators/test_release.py
@@ -29,7 +29,7 @@ def test_normalize_cargo_manifest():
         original_manifest = {
             'dependencies': {
                 'absolute_dep': {'path': '/abs/path/to/dep'},
-                'parent_dep': {'path': '../parent/dep'},
+                'sibling_nested_dep': {'path': '../parent/dep'},
                 'sibling_dep': {'path': '../sibling'},
                 'subdir_dep': {'path': 'subdir/dep'},
                 'dot_subdir_dep': {'path': './subdir/dep'},
@@ -37,11 +37,11 @@ def test_normalize_cargo_manifest():
                 'not_dict_dep': '1.0.0',
             },
             'dev-dependencies': {
-                'dev_parent_dep': {'path': '../dev-dep'},
+                'dev_sibling_dep': {'path': '../dev-dep'},
                 'dev_subdir_dep': {'path': 'dev-subdir'},
             },
             'build-dependencies': {
-                'build_parent_dep': {'path': '../build-dep'},
+                'build_sibling_dep': {'path': '../build-dep'},
                 'build_subdir_dep': {'path': 'build-subdir'},
             }
         }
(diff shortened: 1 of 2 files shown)


test/unit_tests/test_generators/test_release.py
```diff
@@ -0,0 +1,135 @@
...
+    temp_dir = tempfile.mkdtemp()
+    try:
+        manifest_path = os.path.join(temp_dir, 'Cargo.toml')
+
+        # Define the original manifest content
+        original_manifest = {
+            'dependencies': {
+                'absolute_dep': {'path': '/abs/path/to/dep'},
+                'parent_dep': {'path': '../parent/dep'},
+                'sibling_dep': {'path': '../sibling'},
```
Blast545, line 33: In my mental model, `../sibling` it's a sibling of my current crate, correct, but I don't see why `../parent/dep` it's a parent. Looks like it's a child of another sibling, a nephew? 

Maybe I'm getting something wrong

test/unit_tests/test_generators/test_release.py
```diff
@@ -0,0 +1,135 @@
...
+        # Absolute path - preserved!
+        assert deps['absolute_dep']['path'] == '/abs/path/to/dep'
+        assert 'version' not in deps['absolute_dep']
+
+        # Parent referencing path - dropped and version set to *
+        assert 'path' not in deps['parent_dep']
+        assert deps['parent_dep']['version'] == '*'
+
+        assert 'path' not in deps['sibling_dep']
+        assert deps['sibling_dep']['version'] == '*'
```
Blast545: As per the title description, I thought `parent_dep` and `sibling_dep` were going to have a different behavior

Blast545: Also, if this is the expected behavior, maybe it should be better aligned with the title

Blast545 (commented):
I mainly checked the tests and I'm confused, can you leave us a msg about why this is needed? and how it's failing.

I thought I was aware of the bug as per a previous discussion, but I don't get completely get it from test case covered.

test/unit_tests/test_generators/test_release.py
cottsay: This is maybe just a mismatch in nomenclature between what I asked Gemini to do and how I write the commit description.

All of these paths are from the perspective of a package at it's root directory. We archive the packages from that directory, so anything above it (starting with `../`) is removed by Bloom's package-level processing and will NOT be present when the package is built on the buildfarm.

Maybe we just call these `sibling_dep` and `nested_sibling_dep`. Doesn't really matter, either way they should be removed (and possibly converted to a `version = '*'` constraint).

Blast545: > Maybe we just call these `sibling_dep` and `nested_sibling_dep`

That change sounds good to me then, that renaming would makes sense from the description and matches the test behavior

Blast545 (approved):
LGTM

